PR #14, opened 2019-01-24
Fix the head tracking 

GiulioRomualdi:
- revert the sign in the head inverse kinematics
- fix coupling between virtualizer and Oculus headset in the head retargeting
- fix fake imu signal

This issue wants to be a workaround for https://github.com/dic-iit/element_retargeting-from-human/issues/80.

The PR was successfully tested on the robot, 

cc @S-dafarra @kouroshD

### DIFF
--- a/modules/Oculus_module/src/HeadRetargeting.cpp
+++ b/modules/Oculus_module/src/HeadRetargeting.cpp
@@ -42,7 +42,7 @@ void HeadRetargeting::inverseKinematics(const iDynTree::Rotation& chest_R_head, 
     }
 
     // minus due to the joints structure of the iCub neck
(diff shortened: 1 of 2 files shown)


GiulioRomualdi:
Actually, I'm not sure that it's correct to merge this PR, as discussed with @S-Dafarra, this is just a workaround to fix the problem with the neck joints. Probably we should understand better the reason why we have those problems with the neck joints and fix it with a proper PR.

S-Dafarra:
See https://github.com/dic-iit/component_wholebody-teleoperation/issues/156#issuecomment-461797009 for an explanation on the modifications.

PR #23, opened 2019-07-17
Fix the bug related to the Scope of using Xsens in config file

kouroshD:
With this PR we fix the bug related to using xsens option:
During preperation for [ dott. Paolucci demo ](https://github.com/dic-iit/lab-events-demos/issues/112) , we had the problem to use `devel` branch of `walking-teleoperation` to perform walking-Oculus teleoperation (without using Xsens) sceanrio.
The problem was, when we were starting the teleoperaion `startWalking`, the robot hand were going smoothly toward zero.
Investaging on the problem, I understood the problem was from the way I was passing the parameters to `oculusModule`. `useXsens` parameter is defined under the `GENERAL` scope in config files, however, when passing the parameters, I was not identifying `GENERAL`. Therefore, the configuration was not correctly performed, and during the execeution in `oculusModule` the option `useXsens` was `true`.
Additionally, when investigating on this problem, we understood, the ports `/oculusRetargeting/rightHandPose:o` and `/oculusRetargeting/leftHandPose:o` are empty correctly, i.e., nothing written in these port. However, in the controller, it goes toward joint values of `0`. We should investigate more on the walking-controller side and try to fix this problem.

@fjandrad @DanielePucci

### DIFF
--- a/app/robots/iCubGazeboV2_5/XsensRetargeting.ini
+++ b/app/robots/iCubGazeboV2_5/XsensRetargeting.ini
@@ -19,6 +19,6 @@ human_joint_list_stream  ("neck_roll",       "l_elbow",          "l_shoulder_rol
 
 smoothingTime   0.25
 # The max difference (threshold) of a joint value coming from the human (rad)
-jointDifferenceThreshold 1.5
+jointDifferenceThreshold 0.5
 wholeBodyJointsPort           /HumanStateWrapper/state:i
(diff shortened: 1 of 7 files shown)


kouroshD:
@GiulioRomualdi

GiulioRomualdi:
LGTM we can merge the pr!

kouroshD:
Thanks @GiulioRomualdi

PR #25, opened 2019-10-02
get the human joint list online, update the application files for Xse…

kouroshD:
This PR fixes the following problems:
- [Get the joint names order for Xsens Retargeting online from the message instead of config file #88](https://github.com/dic-iit/element_retargeting-from-human/issues/88)
- Update the port names for the data coming from HDE for Xsens retargeting
- update the application files to run the HDE necessary devices for `humanStateProvider`

### DIFF
--- a/app/robots/iCubGenova04/XsensRetargeting.ini
+++ b/app/robots/iCubGenova04/XsensRetargeting.ini
@@ -3,19 +3,12 @@ samplingTime            0.01
 robot                   icub
 
 remote_control_boards   ("head", "torso", "left_arm", "right_arm")
-# Notice the order of the joint list is not wrong.
+# ROBOT JOINT LIST (Notice the order of the joint list is not wrong)
 # Indeed they are written according to the joint order of the walking-coordinator
 joints_list             ("neck_pitch", "neck_roll", "neck_yaw",
                          "torso_pitch", "torso_roll", "torso_yaw",
                          "l_shoulder_pitch", "l_shoulder_roll", "l_shoulder_yaw", "l_elbow", "l_wrist_prosup", "l_wrist_pitch", "l_wrist_yaw",
                          "r_shoulder_pitch", "r_shoulder_roll", "r_shoulder_yaw", "r_elbow",  "r_wrist_prosup", "r_wrist_pitch", "r_wrist_yaw")
-
-# Notice the order of the human joint list is not wrong.
-# Indeed they are written according to the human joint order of the HDE/HumanStateWrapper
-human_joint_list_stream  ("neck_roll",       "l_elbow",          "l_shoulder_roll",    "r_shoulder_yaw", "r_hip_yaw",  "r_shoulder_pitch", "torso_roll",    "l_ankle_roll",
-                          "l_shoulder_yaw",  "r_ankle_roll",     "torso_pitch",        "l_knee",         "r_elbow",    "l_hip_yaw",        "neck_yaw",      "r_wrist_prosup",
-                          "r_wrist_yaw",     "r_shoulder_roll",  "l_shoulder_pitch",   "l_hip_roll",     "neck_pitch",  "r_ankle_pitch",   "l_wrist_prosup", "l_ankle_pitch",
-                          "r_wrist_pitch",   "r_hip_pitch",      "r_knee",             "l_hip_pitch",    "l_wrist_yaw", "r_hip_roll",      "l_wrist_pitch",  "torso_yaw")
 
 smoothingTime   0.25
 # The max difference (threshold) of a joint value coming from the human (rad)
--- a/app/scripts/dcmWalkingRetargetingXsens.xml
+++ b/app/scripts/dcmWalkingRetargetingXsens.xml
@@ -45,6 +45,19 @@
     <node>icub-virtualizer</node>
   </module>
 
+  <module>
+    <name>yarprobotinterface</name>
+    <parameters>--config TransformServer.xml</parameters>
+    <node>icub30</node>
+  </module>
+
+  <module>
+    <name>yarprobotinterface</name>
+    <parameters>--config RobotStateProvider_iCub.xml</parameters>
+    <node>icub30</node>
+  </module>
+
+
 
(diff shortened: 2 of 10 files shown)


app/robots/iCubGenova04/XsensRetargeting.ini
```diff
@@ -3,20 +3,13 @@ samplingTime            0.01
...
 remote_control_boards   ("head", "torso", "left_arm", "right_arm")
-# Notice the order of the joint list is not wrong.
+# ROBOT JOINT LIST (Notice the order of the joint list is not wrong)
 # Indeed they are written according to the joint order of the walking-coordinator
 joints_list             ("neck_pitch", "neck_roll", "neck_yaw",
                          "torso_pitch", "torso_roll", "torso_yaw",
                          "l_shoulder_pitch", "l_shoulder_roll", "l_shoulder_yaw", "l_elbow", "l_wrist_prosup", "l_wrist_pitch", "l_wrist_yaw",
                          "r_shoulder_pitch", "r_shoulder_roll", "r_shoulder_yaw", "r_elbow",  "r_wrist_prosup", "r_wrist_pitch", "r_wrist_yaw")
 
-# Notice the order of the human joint list is not wrong.
```
GiulioRomualdi: Is this removed because now the name of the joints are streamed from the wearable device`?

app/scripts/dcmWalkingRetargetingXsens.xml
```diff
@@ -45,6 +45,19 @@
...
 
+  <module>
+    <name>yarprobotinterface</name>
+    <parameters>--config TransformServer.xml</parameters>
+    <node>icub30</node>
+  </module>
+
+  <module>
+    <name>yarprobotinterface</name>
+    <parameters>--config RobotStateProvider_iCub.xml</parameters>
```
GiulioRomualdi: Where `RobotStateProvider_iCub.xml` can be found? Is it provided by another repo? If yes you should add the dependences in the readme.

modules/Xsens_module/src/XsensRetargeting.cpp
```diff
@@ -3,7 +3,16 @@
 #include <iterator>
 #include <sstream>
 
-XsensRetargeting::XsensRetargeting(){};
+class XsensRetargeting::impl
+{
+public:
+    bool mapJointsHDE2CONTROLLER(std::vector<std::string> robotJointsListNames,
```
GiulioRomualdi: Can you add the documentation related to this function? Furthermore is there a reason why `mapJointsHDE2CONTROLLER` is not called `mapJointsHDE2Controller`?

modules/Xsens_module/src/XsensRetargeting.cpp
```diff
@@ -179,8 +127,9 @@ bool XsensRetargeting::getJointValues()
 
     // in the msg thrift the first list is the joint names [get(0)], second the joint values
     // [get(1)]
-    yarp::os::Value humanjointsValueS = desiredHumanJoints->get(1);
-    yarp::os::Bottle* tmpHumanNewJointValues = humanjointsValueS.asList();
+
+    yarp::os::Value humanjointsValues = desiredHumanJoints->get(1);
+    yarp::os::Bottle* tmpHumanNewJointValues = humanjointsValues.asList();
```
GiulioRomualdi: If the message is autogenerated from a thrift, you can import the library generated by Yarp and avoid to use the bottles.

modules/Xsens_module/src/XsensRetargeting.cpp
```diff
@@ -258,3 +254,43 @@ bool XsensRetargeting::close()
...
+{
+    if (!humanToRobotMap.empty())
+    {
+        humanToRobotMap.clear();
+    }
+
+    bool foundMatch = false;
+    for (unsigned i = 0; i < robotJointsListNames.size(); i++)
+    {
+        for (unsigned j = 0; j < humanJointsListName.size(); j++)
```
GiulioRomualdi: probably you can simplify this part of the code using [`std::find`](https://en.cppreference.com/w/cpp/algorithm/find) and [`std::distance`](https://en.cppreference.com/w/cpp/iterator/distance)

app/robots/iCubGenova04/XsensRetargeting.ini
kouroshD: yes.

DanielePucci:
@kouroshD any update on this?

modules/Xsens_module/src/XsensRetargeting.cpp
kouroshD: Thank you for the suggestion. If I add the dependency to the autogenerated library, it means `teleoperation` profile at build-time will have dependency to `human-dynamics` profile, which I would like to avoid it.

modules/Xsens_module/src/XsensRetargeting.cpp
kouroshD: Using these standard commands are an option as well. However, I do not see any simplifications using them. Using them, we still will have one `for` loop, and inside that we should iterate using `find` and `if` condition, then using `distance` to check the element number. The inner loop is doing exactly the same! From the computational point of view, both approaches are linear in time.

modules/Xsens_module/src/XsensRetargeting.cpp
kouroshD: updated.

kouroshD:
During the sprint https://github.com/dic-iit/component_wholebody-teleoperation/issues/177, we were using the open PR branches, and it was working as expected. When all the comments are done, we can merge it.

modules/Xsens_module/src/XsensRetargeting.cpp
GiulioRomualdi: ok!

modules/Xsens_module/src/XsensRetargeting.cpp
GiulioRomualdi: Yes it is fine for me

PR #35, opened 2020-06-30
Merge Devel to Master

kouroshD:

PR #36, opened 2020-07-03
Merge pull request #35 from robotology/devel

kouroshD:
Merge Master to devel after this PR https://github.com/robotology/walking-teleoperation/pull/35


kouroshD:
> For commits with no change, you can directly push 😉

Following the @S-Dafarra  suggestion, I pushed directly to avoid making a new merge commit.
So, I close this PR.

PR #37, opened 2020-07-03
Update the Virtualizer Module with the new version of the Cyberith treadmill 

kouroshD:
with this PR, we update the virtualizer module with the new version of the cyberith treadmill SDK.
Moreover, remove the old file of the virtulizer sdk.


cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,67 @@
...
 
-# Check Directory Macro
-macro(CHECK_DIR _DIR)
-  if(NOT EXISTS "${${_DIR}}")
-    message(WARNING "Directory \"${${_DIR}}\" not found.")
-    set(CybSDK_FOUND FALSE)
-    unset(_DIR)
-  endif()
-endmacro()
+if(NOT DEFINED ENV{CybSDK_INCLUDE_DIRS})
```
S-Dafarra, line 20: I would suggest not requesting the definition of multiple variables from the user. If the structure on the SDK is known, we can just "hardcode" the relative paths starting from  ``CybSDK_DIR``

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,67 @@
...
 
-# Target Platform
-set(TARGET_PLATFORM)
-if(NOT CMAKE_CL_64)
-  set(TARGET_PLATFORM x86)
+if(WIN32)
+  find_library(CybSDK_LIBRARY NAMES CybSDK HINTS ${CybSDK_DIR})
 else()
-  set(TARGET_PLATFORM x64)
+  message(FATAL_ERROR  "The Sense Glove SDK is not supported for this operating system.")
```
S-Dafarra, line 32: Copy-paste leftover?

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,67 @@
...
 # This will define the following variables::
 #
-#   CybSDK_FOUND    - True if the system has the CybSDK 
-#   CybSDK_VERSION  - The version of the CybSDK  which was found
-#
-#
+#   CybSDK_FOUND         - True if the system has the Sense Glove SDK
+#   CybSDK_INCLUDE_DIR  - The include directory of the SDK
+#   CybSDK_LIBRARIES     - The name of the libraries to link against.
+###############################
```
S-Dafarra, line 11: In modern CMake, it is better to define targets. Recently, I wrote these lines for a similar purpose (need to be tested)
```cmake
    if (BUILD_SHARED_LIBS)
        set(EXTENSION ${CMAKE_SHARED_LIBRARY_SUFFIX})
        set(TYPE "SHARED")
    else()
        set(EXTENSION ${CMAKE_STATIC_LIBRARY_SUFFIX})
        set(TYPE "STATIC")
    endif()

    if(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
        # 64 bits
        set(FOLDER "x64")
    elseif(${CMAKE_SIZEOF_VOID_P} EQUAL 4)
        # 32 bits
        set(FOLDER "x86")
    endif()

    add_library(CybSDK ${TYPE} IMPORTED GLOBAL ${CybSDK_DIR}/${FOLDER}/CybSDK${EXTENSION})
    set_target_properties(CybSDK PROPERTIES IMPORTED_LOCATION ${CybSDK_DIR}/${FOLDER}/CybSDK${EXTENSION})
    target_include_directories(CybSDK INTERFACE ${CybSDK_DIR}/include)
```

The first two ``if``s are only needed to determine the type of library to link. The last three lines are the important ones. The first one generates the imported target, the second embeds in the target the location of the library, the third define from where to take the ``.h`` files. Then, you can simply use this target in other components by doing
```cmake
target_link_libraries(targetName CybSDK)
```
without changing anything else.

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,67 @@
...
+
+# Include Directories
+#set(CybSDK_INCLUDE_DIRS)
+#if(CybSDK_FOUND)
+#  set(CybSDK_INCLUDE_DIRS $ENV{CybSDK_DIR}/Core/SGCoreCpp/incl)
+#endif()
+
+#if(NOT (EXISTS CybSDK_INCLUDE_DIRS))
+#  message(FATAL_ERROR "CybSDK_INCLUDE_DIRS not exists: ${CybSDK_INCLUDE_DIRS}")
+#endif()
```
S-Dafarra, line 65: Please avoid leaving commented code around. It may make the user wonder if those lines are actually useful or not. Same comment for the other commented code.

modules/Virtualizer_module/src/VirtualizerModule.cpp
```diff
@@ -206,9 +206,10 @@ bool VirtualizerModule::updateModule()
 
     double tmpSpeedDirection = (double)(m_cvirtDeviceID->GetMovementDirection());
     double speedDirection = 1.0; // set the speed direction to forward by default.
+    
     if (std::abs(tmpSpeedDirection) < 0.01)
         speedDirection = 1.0;
-    else if (std::abs(tmpSpeedDirection - 1) < 0.01)
+    else if (std::abs(tmpSpeedDirection + 1) < 0.01)
```
S-Dafarra, line 212: Why this change? And what is this ``1``?

kouroshD: Basically, in the first version of the sdk, if the person was walking backward in the sdk the output value from the sdk indicating the direction of the motion was `+1`. In the new version of the SDK, this value is `-1` (which makes more sense actually)!
I'll add a comment on top of it explaining this value.

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,67 @@
...
-#
-#
+#   CybSDK_FOUND         - True if the system has the Sense Glove SDK
+#   CybSDK_INCLUDE_DIR  - The include directory of the SDK
+#   CybSDK_LIBRARIES     - The name of the libraries to link against.
+###############################
 
-##### Utility #####
+# Check Directory of the CybSDK_DIR
+if(NOT DEFINED ENV{CybSDK_DIR})
```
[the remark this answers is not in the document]
kouroshD: > Note that this variable cannot be set from the ccmake interface.

You mean 'can', am I right?
Basically, you are saying that I create an option for with a cmake variable, and initialize it with the environment variable. Later, the user can update the value of it if he wants!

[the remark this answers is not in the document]
S-Dafarra: > You mean 'can', am I right?

No no, I meant what I wrote. You cannot change or visualize that variable from ``ccmake``. Nevertheless, you fully understood my point, i.e
> Basically, you are saying that I create an option for with a cmake variable, and initialize it with the environment variable. Later, the user can update the value of it if he wants!

modules/Virtualizer_module/src/VirtualizerModule.cpp
S-Dafarra: Ok, then if the value is negative, isn't it sufficient to check directly the sign of ``tmpSpeedDirection``?

kouroshD: No, since the sdk method is returning a float variable, I wanted to be sure if the value we are receiving is correct. For example, in case for some reason the variable is not set or initialized correctly.

kouroshD: I have added the commnet.

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,34 @@
...
-  set(TARGET_PLATFORM x86)
-else()
-  set(TARGET_PLATFORM x64)
+set(CybSDK_DIR $ENV{CybSDK_DIR} CACHE STRING "The directiry containing the Cyberith library")
+if(NOT DEFINED CybSDK_DIR)
+  message( FATAL_ERROR "variable {CybSDK_DIR} is not defined: ${CybSDK_DIR}" )
 endif()
 
-##### Find CybSDK #####
+option (BUILD_SHARED_LIBS "Please identify the library type SHARED or STATIC" FALSE)
```
S-Dafarra, line 11: This line should go the main ``CMakeLists.txt`` if not already present.

modules/Virtualizer_module/CMakeLists.txt
```diff
@@ -44,8 +38,7 @@ yarp_add_idl(${EXE_TARGET_NAME}_THRIFT_SRC ${${EXE_TARGET_NAME}_THRIFT_HDR})
 
 # add include directories to the build.
 include_directories(
-  ${CMAKE_CURRENT_SOURCE_DIR}/include
-  ${CMAKE_CURRENT_SOURCE_DIR}/CybSDK/include)
+  ${CMAKE_CURRENT_SOURCE_DIR}/include )
```
S-Dafarra, line 41: You can substitute this with [``target_include_directories``](https://cmake.org/cmake/help/v3.3/command/target_include_directories.html). The difference is that you are more specific in which target that folder should be used. Even if for executables this does not change anything, for libraries it is more important as it determines where headers can be found.

modules/Virtualizer_module/CMakeLists.txt
```diff
@@ -44,8 +38,7 @@ yarp_add_idl(${EXE_TARGET_NAME}_THRIFT_SRC ${${EXE_TARGET_NAME}_THRIFT_HDR})
 
 # add include directories to the build.
 include_directories(
-  ${CMAKE_CURRENT_SOURCE_DIR}/include
-  ${CMAKE_CURRENT_SOURCE_DIR}/CybSDK/include)
+  ${CMAKE_CURRENT_SOURCE_DIR}/include )
 
 # add an executable to the project using the specified source files.
 add_executable(${EXE_TARGET_NAME} ${${EXE_TARGET_NAME}_SRC} ${${EXE_TARGET_NAME}_HDR}
```
S-Dafarra, line 44: Probably a 
```cmake
target_link_libraries(${EXE_TARGET_NAME} PUBLIC CybSDK)
```
is missing.

modules/Virtualizer_module/CMakeLists.txt
```diff
@@ -57,4 +50,4 @@ target_link_libraries(${EXE_TARGET_NAME}
   UtilityLibrary)
 
 install(TARGETS ${EXE_TARGET_NAME} DESTINATION bin)
-install(FILES ${CybSDK_LIBRARIES} DESTINATION bin)
+#install(FILES CybSDK DESTINATION bin)
```
S-Dafarra, line 53: Please avoid leaving commented code around.

modules/Virtualizer_module/src/VirtualizerModule.cpp
S-Dafarra: > No, since the sdk method is returning a float variable, I wanted to be sure if the value we are receiving is correct. For example, in case for some reason the variable is not set or initialized correctly.

To be honest, this is not really clear to me. Why is it different from doing
```c++
if (tmpSpeedDirection < -0.99)
```
? I usually prefer having magic numbers around. Even if I get the meaning, the reason of having that ``+1`` is not immediately clear to me.

kouroshD: there is the difference, if you check there is the distance from `-1` so it clearly different from `if (tmpSpeedDirection < -0.99)`.With the comment of the line, should be cristal clear now.

cmake/FindCybSDK.cmake
kouroshD: I am not sure, since if we add this line to the main CMakeLists, then we are setting this for all the libraries. In this case, it is only for this library.

modules/Virtualizer_module/CMakeLists.txt
kouroshD: There is already!

https://github.com/kouroshD/walking-teleoperation/blob/d4cc72ad8bca495c84e1010d00d9ec5a607739e1/modules/Virtualizer_module/CMakeLists.txt#L47-L50

cmake/FindCybSDK.cmake
S-Dafarra: That's not an issue. Usually, you may want an entire project to be built statically, not just some libraries.

modules/Virtualizer_module/src/VirtualizerModule.cpp
S-Dafarra: Yeah I got meaning, but I am not understanding why you are checking it in this way. In my opinion that condition can be simplified to be more direct, but anyway it is not blocking.

modules/Virtualizer_module/CMakeLists.txt
kouroshD: done!

modules/Virtualizer_module/src/VirtualizerModule.cpp
kouroshD: Which part of it you did not understand?!

cmake/FindCybSDK.cmake
kouroshD: yes, but what if we want we have some shared libraries only that we want to include it to the executable?
btw, by default this is False, so, all the libraries are built static.

modules/Virtualizer_module/src/VirtualizerModule.cpp
S-Dafarra: We quickly discussed F2F. What I did not understand was the ``abs`` was used as a distance metric from the point ``-1``. The condition is valid only if the  ``tmpSpeedDirection`` is in the range ``[-1.01, -0.99]``.

cmake/FindCybSDK.cmake
S-Dafarra: We quickly discussed F2F and a possible solution is to have a dedicated option to select which version to link. Here my suggestion:
```cmake
option (LINK_CYBSDK_SHARED "Please identify the if the CybSDK library should be SHARED or STATIC" ${BUILD_SHARED_LIBS})
```

S-Dafarra:
The actions are failing because before the SDK was shipped with the library. In addition, as discussed F2F this line https://github.com/robotology/walking-teleoperation/blob/d63a19cefc416cae4dd6b195c03bd25c917f8e92/cmake/FindCybSDK.cmake#L7 is not working as expected because the variable is set, but to a non-existent path. It can be changed to
```
if(NOT EXISTS CybSDK_DIR) 
```
If the ``CybSDK`` is not found, the module should not be compiled, or there should be an option which avoids compiling the module and looking for the SDK.

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,38 @@
...
+    set(FOLDER "x86")
 endif()
+##### Find CybSDK #####
 
-message(STATUS "CybSDK_FOUND : ${CybSDK_LIBRARIES}")
+add_library(CybSDK ${TYPE} IMPORTED GLOBAL ${CybSDK_DIR}/${FOLDER}/CybSDK${EXTENSION})
+set_target_properties(CybSDK PROPERTIES IMPORTED_LOCATION ${CybSDK_DIR}/${FOLDER}/CybSDK${EXTENSION})
+target_include_directories(CybSDK INTERFACE ${CybSDK_DIR}/Include)
 
+set(CybSDK_FOUND TRUE)
```
S-Dafarra, line 38: Remember to add a terminal blank line at the end. This avoids problems in some systems.

modules/Virtualizer_module/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@
 #
 # set target name
 
-find_package(CybSDK REQUIRED)
 set(EXE_TARGET_NAME VirtualizerModule)
 
+
+find_package(CybSDK REQUIRED)
```
S-Dafarra, line 11: Probably you should put ``QUIET`` instead of ``REQUIRED``. When a ``REQUIRED`` package is not found, ``cmake`` should exit with an error.

cmake/FindCybSDK.cmake
```diff
@@ -1,86 +1,38 @@
...
-  set(CybSDK_FOUND FALSE)
-endif()
-
-# Root Directoty
-set(CybSDK_DIR)
-if(CybSDK_FOUND)
-  set(CybSDK_DIR $ENV{CybSDK_DIR} CACHE PATH "CybSDK Path." FORCE)
-  check_dir(CybSDK_DIR)
-endif()
+option (BUILD_SHARED_CybSDK_LIB "Identify if the library type is SHARED or STATIC" FALSE)
```
S-Dafarra, line 13: Not really a problem, but you can initialize this variable with ${BUILD_SHARED_LIBS}.

modules/Virtualizer_module/CMakeLists.txt
```diff
@@ -17,14 +24,8 @@ add_install_rpath_support(BIN_DIRS "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BIND
   DEPENDS ENABLE_RPATH
   USE_LINK_PATH)
 
-# find virtualizer SDK
-file(GLOB CybSDK_SOURCE_H "${CMAKE_CURRENT_SOURCE_DIR}/CybSDK/include/*")
-file(GLOB CybSDK_SOURCE_C "${CMAKE_CURRENT_SOURCE_DIR}/CybSDK/src/*")
-
 add_definitions(-DUNICODE)
```
S-Dafarra, line 27: Probably this is not needed anymore. If it is needed, you can use [``target_compile_definitions``](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html)

modules/Virtualizer_module/CMakeLists.txt
kouroshD: No, I have tested both when found and not found. it is working as it should be. btw, the library is required, so I am not sure if putting quiet is fine.

cmake/FindCybSDK.cmake
kouroshD: No, I prefer to have it with static type as the default value of the option for simplification. If you have any strong reason for changing that, let me know.

cmake/FindCybSDK.cmake
kouroshD: I am not getting, are you telling me that to add a blank line before or/and after this line!?

modules/Virtualizer_module/CMakeLists.txt
S-Dafarra: ``QUIET`` means that ``cmake`` does not crash and burn if the package is not found and no messages are printed. Nevertheless, you can still have some ``cmake`` machinery later on to complain if you really wanted that dependency.

cmake/FindCybSDK.cmake
S-Dafarra: A blank line at the end of the file. The github interface shows this with a red symbol
image

modules/Virtualizer_module/CMakeLists.txt
kouroshD: yes, we have also later. but if it is needed for sure this dependency, then it is fine to burn and crash!

S-Dafarra: But then the ``if`` later does not make much sense https://github.com/robotology/walking-teleoperation/blob/70a73b46c5101aaaaedbf3dd9e6c2d5e652a5ea3/modules/Virtualizer_module/CMakeLists.txt#L11-L12 since either is always ``true`` or it should not be reached at all because ``cmake`` should exit with an error. I guess that right now ``REQUIRED`` is not used at all in the ``FindPackage``. I discovered later that you can use ``find_package_handle_standard_args`` to make your custom ``FIndPackage`` behave like a classic one. It is the one throwing errors when the library is not found. You can see an example of usage here https://github.com/iit-danieli-joint-lab/idjl-software/blob/master/cmake/extern/FindMATIO.cmake#L22

Nevertheless, this is not fundamental at the moment, even if it would be cleaner.

S-Dafarra:
Just to bother you a little more :grin:, for the future try to use a more detailed commit message. This may all go in the repo history and it may be hard to understand later on what this meant
image

kouroshD:
> Just to bother you a little more 😁, for the future try to use a more detailed commit message. This may all go in the repo history and it may be hard to understand later on what this meant
> image

I normally do that, if there are some "messages" to be conveyed in the commit. I am going to squash the PR before merging.

kouroshD:
@GiulioRomualdi  @traversaro  If you are OK with the PR, I merged it.

traversaro:
@kouroshD Thanks for asking, but for the future I think one reviewer is enough.

PR #42, opened 2020-11-17
add glove option to the Oculus Retargeting Module

kouroshD:
With this PR, we add the option in the oculus module to have sense Glove.
In the config files, at the moment, this is only done for Teleoperation scenario with Xsens. We can also add the option for oculus teleoperation setup later, if it works.


S-Dafarra (approved):

kouroshD:
Thanks for the review. Merging.

PR #47, opened 2021-01-12
Add configuration file for XPRIZE demo

GiulioRomualdi:
I would test it on the real setup before merging the PR


app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,217 @@
...
+
+  <!-- Modules -->
+  <!-- icub-head -->
+
+  <module>
+    <name>yarpdev</name>
+    <parameters> --from camera/ServerGrabberDualDragon.ini --split true</parameters>
+    <node>icub-head</node>
+  </module>
+
```
prashanthr05, line 22: Add the module related to the transform server.

```
<module>
    <name>yarpdev</name>
    <parameters> --device transformServer --ROS::enable_publisher 1 --ROS::enable_subscriber 0</parameters>
    <node>icub-console-gui</node>
  </module>
```

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,217 @@
...
+  <module>
+    <name>yarpdev</name>
+    <parameters> --device AudioRecorderWrapper --subdevice portaudioRecorder --name /icub/microphone  --min_samples_over_network 4000  --max-samples_over_network 4000  --rate 160000 --samples 16000</parameters>
+    <node>icub-head</node>
+  </module>
+
+  <module>
+    <name>yarpdev</name>
+    <parameters>--device AudioPlayerWrapper --subdevice portaudioPlayer --name /icub/speakers --playback_network_buffer_lenght 0.4</parameters>
```
prashanthr05, line 37: Need to document the parameter renaming in the newer versions of ROS (post November 2020) 
`playback_network_buffer_lenght`  -> `playback_network_buffer_size`.

See https://github.com/robotology/yarp/commit/1734f53a4bbc30755b7819fae2b4be803a3a6bec

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,217 @@
...
+    <dependencies>
+      <port timeout="5.0" request="start" reply="[ok]">/icub-virtualizer/microphone/rpc</port>
+    </dependencies>
+  </module>
+
+  <!-- icub-console -->
+
+  <module>
+    <name>camCalibWithPose</name>
+    <parameters>--context cameraCalibration --from icubEyes.ini --group CAMERA_CALIBRATION_LEFT --name /icub/camcalib/left --useIMU</parameters>
```
prashanthr05, line 48: Need to check how the `--useIMU` option affects running the module on the robots equipped with IMUs streaming through MAS interface, for example `iCubGenova09`.

app/robots/iCubGenova09/oculusConfig.ini
```diff
@@ -0,0 +1,54 @@
...
+playerOrientationPort   /playerOrientation:i
+rpcWalkingPort_name     /walkingRpc
+rpcVirtualizerPort_name /virtualizerRpc
+
+[GENERAL]
+samplingTime            0.01
+robot                   icub
+useXsens                1
+enableLogger            0
+enableMoveRobot         0
```
S-Dafarra, line 15:
```suggestion
enableMoveRobot         1
```

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,217 @@
...
+  <module>
+    <name>camCalibWithPose</name>
+    <parameters>--context cameraCalibration --from icubEyes.ini --group CAMERA_CALIBRATION_RIGHT --name /icub/camcalib/right --useIMU</parameters>
+    <node>icub-console</node>
+  </module>
+
+  <!-- icub-virtualizer -->
+  <module>
+    <name>yarpdev</name>
+    <parameters> --device JoypadControlServer --use_separate_ports 1 --period 10 --name /joypadDevice/Oculus --subdevice ovrheadset --sticks 0  --tfDevice transformClient --tfLocal /oculustf --tfRemote /transformServer --tf_head_frame headoculus --tf_left_hand_frame loculus --tf_right_hand_frame roculus --tf_root_frame oculusworld --stick_as_axis true --gui_elements 0</parameters>
```
S-Dafarra, line 61:
```suggestion
    <parameters> --device JoypadControlServer --use_separate_ports 1 --period 10 --name /joypadDevice/Oculus --subdevice ovrheadset --sticks 0  --tfDevice transformClient --tfLocal /oculustf --tfRemote /transformServer --tf_head_frame headoculus --tf_left_hand_frame loculus --tf_right_hand_frame roculus --tf_root_frame oculusworld --stick_as_axis true --gui_elements 0 --userpose true</parameters>
```

S-Dafarra: This is to avoid the images to be moving in the headset

S-Dafarra:
Probably we should also enable the automatic calibration of the cameras at startup.

app/scripts/xprizeDemo.xml
prashanthr05: Apparently using `useIMU` option with `iCubGenova09` does not work. On removing that option, the module starts streaming the images.

lrapetti:
We should also add the Xsens retargeting related devices plus the transform server

S-Dafarra:
> We should also add the Xsens retargeting related devices plus the transform server

Actually, it seems to be better to have them in a separate application as it is right now. In this way, stopping all the applications will not stop the Xsens visualization.

app/scripts/xprizeDemo.xml
S-Dafarra: See https://github.com/robotology/walking-teleoperation/pull/47#issuecomment-767554588

S-Dafarra (approved):
We successfully tested it on the real robot.

kouroshD:
@GiulioRomualdi  is the PR ready for review? if yes, I will do it on Monday.

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,243 @@
...
+     <node>icub-console-gui</node>
+   </module>
+
+  <module>
+    <name>HapticGloveModule</name>
+    <node>icub-console-gui</node>
+  </module>
+
+  <module>
+    <name>dyn-visualizer</name>
```
kouroshD, line 111: Is this module located at iDynTree? About its installation, is it installed by default or needs some options to be enabled?
I found it here https://github.com/S-Dafarra/dyn-visualizer . If it is the one we used, we may add it move it to DIC or Robotology organizations.

kouroshD:
@GiulioRomualdi  @S-Dafarra  I think, we may add a document in the docs folder in order to documents the modules and dependencies to install. Some codes are not yet merged and may some time to merge them. For this reason, at least we have a documentation of them.

S-Dafarra:
@kouroshD both the ``dyn-visualizer`` and ``face-expressions-from-mic`` are two applications that will be merged in this repo. To be honest I would avoid spending time writing documentation for those two modules as they are supposed to be moved soon.

In any case, they are currently located in https://github.com/S-Dafarra/face_expressions_from_mic and https://github.com/S-Dafarra/dyn-visualizer. All the configurations are hardcoded. When moving them inside this repo I will also take care of modifying the application, but to be honest, I would consider this not blocking.

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,243 @@
...
+  <module>
+    <name>camCalibWithPose</name>
+    <parameters>--context cameraCalibration --from icubEyes.ini --group CAMERA_CALIBRATION_RIGHT --name /icub/camcalib/right</parameters>
+    <node>icub-console</node>
+  </module>
+
+  <!-- icub-virtualizer -->
+  <module>
+    <name>yarpdev</name>
+    <parameters> --device JoypadControlServer --use_separate_ports 1 --period 10 --name /joypadDevice/Oculus --subdevice ovrheadset --sticks 0  --tfDevice transformClient --tfLocal /oculustf --tfRemote /transformServer --tf_head_frame headoculus --tf_left_hand_frame loculus --tf_right_hand_frame roculus --tf_root_frame oculusworld --stick_as_axis true --userpose true --gui_elements 1 --GUI_0::width 0.15 --GUI_0::height 0.15 --GUI_0::x 0.2 --GUI_0::y -0.2 --GUI_0::z -0.5 --GUI_0::alpha 1.0 --hfov 59 --left-pitch-offset 0.0725 --left-yaw-offset -0.065 --right-pitch-offset 0.13 --right-yaw-offset 0.07</parameters>
```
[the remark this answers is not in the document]
S-Dafarra: This is the result of several investigations and testings. The related issues are in private repos and I think it is not worth to add them here. For the sake of clarity:
- ``--userpose true`` locks the cameras in front of the operators' eyes.
- ``--gui_elements 1 --GUI_0::width 0.15 --GUI_0::height 0.15 --GUI_0::x 0.2 --GUI_0::y -0.2 --GUI_0::z -0.5 --GUI_0::alpha 1.0`` are needed to have the robot visualization in the oculus (I actually got this by looking inside the device code)
- ``--hfov 59 --left-pitch-offset 0.0725 --left-yaw-offset -0.065 --right-pitch-offset 0.13 --right-yaw-offset 0.07`` are a set of parameters to change the position of the two cameras in the headset. We tuned it according to the ``iCub3`` eyes.

S-Dafarra:
I think we should consider the ``xprizeDemo.xml`` related to iCub3 only btw, both for the definition of the nodes, but also for all the setup (like the oculus parameters).

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,243 @@
...
+   <module>
+     <name>XsensRetargetingModule</name>
+     <node>icub-console-gui</node>
+   </module>
+
+  <module>
+    <name>HapticGloveModule</name>
+    <node>icub-console-gui</node>
+  </module>
+
```
S-Dafarra, line 109:
```suggestion
<!-- This module can be preliminarily found at https://github.com/S-Dafarra/dyn-visualizer -->
```

app/scripts/xprizeDemo.xml
```diff
@@ -0,0 +1,243 @@
...
+  <module>
+    <name>dyn-visualizer</name>
+    <node>icub-console-gui</node>
+  </module>
+
```
S-Dafarra, line 114:
```suggestion
<!-- This module can be preliminarily found at https://github.com/S-Dafarra/face_expressions_from_mic -->
```

S-Dafarra (commented):
@kouroshD I have just added two suggestions for finding the modules, as discussed F2F. Feel free to accept them, or to modify them

app/virtualizerRetargeting/virtualizerConfig.ini
```diff
@@ -8,5 +8,5 @@ deadzone        0.35
 playerOrientationPort_name    /playerOrientation:o
 robotOrientationPort_name     /robotOrientation:i
 rpcWalkingPort_name           /walkingRpc
-scale_X                       2.0
-scale_Y                       1.0
+scale_X                       3.0
+scale_Y                       2.0
```
kouroshD, line 12: Why here, we changed the `scale_Y`? As you can see here in the following line of the code:
https://github.com/robotology/walking-teleoperation/blob/9bc40b0f6e301362a5ac4b7fd0c14bb52b21b744/modules/Virtualizer_module/src/VirtualizerModule.cpp#L225
By setting this value to 2, actually if the user rotate 90 degrees inside the virtualizer, the robot will rotate about 180 degrees. At least, this value will be provided to the unicycle planner.

kouroshD: This also explains why when we were teleoperating the robot with @GiulioRomualdi, the robot was rotating a lot and it was hard to control the rotational motion of the robot teleoperated.

kouroshD (commented):
> I think we should consider the `xprizeDemo.xml` related to iCub3 only btw, both for the definition of the nodes, but also for all the setup (like the oculus parameters).

Thatd's true, but for Xprize and icub3, I expect to have eventually several types of teleoperation to have different plans when teleoperating the robot. And, with a more specific naming it is clear the Xprize demo. That's why I suggested being more specific. This can be reached even with versioning numbers, which I found it less intuitive.

app/virtualizerRetargeting/virtualizerConfig.ini
kouroshD: Talking with @S-Dafarra , it came out that the increasing the  scaling factor does not mean two times the amount of rotation, and we both agree that the system will more reactive to the user.

As we tested several times before on iCub 2.5, the value `1`, here also I suggest keeping the value `1` as the default value.

S-Dafarra: > As we tested several times before on iCub 2.5, the value `1`, here also I suggest keeping the value `1` as the default value.

It is a number that after some trials seemed to be working better. It is probably a little low as well, considering that we are walking at a speed slower than what we achieve with the joystick.

app/robots/iCubGenova09/leftHandRetargetingParams.ini
```diff
@@ -0,0 +1,17 @@
+# additional rotations
+# the following rotation map the hand oculus frame and the hand robot frame (Identity)
+handOculusFrame_R_handRobotFrame ((0.0 0.0 -1.0), (-1.0 0.0 0.0), (0.0 1.0 0.0))
```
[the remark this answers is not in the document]
kouroshD: As mentioned by @S-Dafarra , the values are obtained from the robot frames and tried on the robot. So, we can resolve this point.

app/robots/iCubGenova09/oculusConfig.ini
```diff
@@ -0,0 +1,60 @@
...
+enableLogger            0
+enableMoveRobot         1
+# the following value is a threshold used to update the teleoperation frame position
+# when the human rotates inside the virtualizer
+playerOrientationThreshold    0.2
+
+# For kinematic scaling for task-space retargeting
+humanHeight                   1.76
+# The robot arm span is the distance between the left and right index fingertips when the arms are aligned, i.e. T pose.
+robotArmSpan                  1.0
```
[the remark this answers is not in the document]
kouroshD: As mentioned by @S-Dafarra in f2f discussion, the values are obtained from the trials with the robot. It may need further tuning and checks in the future.

app/robots/iCubGenova09/rightHandRetargetingParams.ini
```diff
@@ -0,0 +1,17 @@
+# additional rotations
+# the following rotation map the hand oculus frame and the hand robot frame (Identity)
+handOculusFrame_R_handRobotFrame ((0.0 0.0 -1.0), (-1.0 0.0 0.0), (0.0 1.0 0.0))
```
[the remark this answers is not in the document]
kouroshD: As mentioned by @S-Dafarra , the values are obtained from the robot frames and tried on the robot. So, we can resolve this point.

app/robots/iCubGenova09/leftHandRetargetingParams.ini
```diff
@@ -0,0 +1,17 @@
...
+# The z-axis is parallel to gravity but pointing upwards.
+# The x-axis points behind the robot.
+# The y-axis points laterally and is chosen according to the right-hand rule.
+# the Teleoperation frame is attached to the virtualizer and the  origin is the
+# same oculus inertial frame. In other words:
+# The z-axis is parallel to gravity but pointing upwards
+# The x-axis points forward
+# The y-axis points laterally and is chosen according to the right-hand rule.
+# So the rotation matrix is simply given by Rotz(pi)
+teleoperationRobotFrame_R_teleoperationFrame ((1.0 0.0 0.0), (0.0 1.0 0.0), (0.0 0.0 1.0))
```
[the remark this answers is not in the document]
kouroshD: As mentioned by @S-Dafarra , the values are obtained from the robot frames and tried on the robot. So, we can resolve this point.

app/robots/iCubGenova09/rightHandRetargetingParams.ini
```diff
@@ -0,0 +1,17 @@
...
+# The z-axis is parallel to gravity but pointing upwards.
+# The x-axis points behind the robot.
+# The y-axis points laterally and is chosen according to the right-hand rule.
+# the Teleoperation frame is attached to the virtualizer and the  origin is the
+# same oculus inertial frame. In other words:
+# The z-axis is parallel to gravity but pointing upwards
+# The x-axis points forward
+# The y-axis points laterally and is chosen according to the right-hand rule.
+# So the rotation matrix is simply given by Rotz(pi)
+teleoperationRobotFrame_R_teleoperationFrame ((1.0 0.0 0.0), (0.0 1.0 0.0), (0.0 0.0 1.0))
```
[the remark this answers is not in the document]
kouroshD: As mentioned by @S-Dafarra , the values are obtained from the robot frames and tried on the robot. So, we can resolve this point.

app/virtualizerRetargeting/virtualizerConfig.ini
kouroshD: Following the f2f discussion, since we did the experiment with the value in this config file, and we keep it for now here, Later, a tuning session can be done to check this parameter.

app/scripts/xprizeDemo.xml
kouroshD: We will open an issue for this point, and in future when we do update the iCub3 head, we will update this file here.

kouroshD (approved):

PR #55, opened 2021-03-18
Add face expressions retargeting module

S-Dafarra:
This PR adds a module to control the facial expressions when it detects some voice through the microphone.

It also cleans a bit the CMake machinery.

### DIFF
--- a/modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
+++ b/modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
@@ -1,0 +1,181 @@
+/**
+ * @file FaceExpressionsRetargeting.cpp
+ * @authors Stefano Dafarra <stefano.dafarra@iit.it>
+ * @copyright 2021 iCub Facility - Istituto Italiano di Tecnologia
+ *            Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ * @date 2021
+ */
+
+#include <FaceExpressionsRetargeting.hpp>
+#include <yarp/os/Vocab.h>
+#include <yarp/os/LogStream.h>
+#include <functional>
+#include <cmath>
+
+#define VAD_FREQUENCY     8000 //Other frequencies do not seem to work well
+#define VAD_SAMPLE_LENGTH 80 //This seeemed to be the only value for which there was a usable output.
+
+bool FaceExpressionsRetargetingModule::configure(yarp::os::ResourceFinder &rf)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    yInfo() << "Configuring";
+    m_fvadObject = fvad_new();
+
+    if (!m_fvadObject)
+    {
+        yError() << "Failed to created VAD object";
+        return false;
+    }
+
+    std::string name = rf.check("name", yarp::os::Value("face_expressions_from_mic"), "The name of the module.").asString();
+    setName(name.c_str());
+
+    std::string audioPortIn = rf.check("audio_input_port_name", yarp::os::Value("/in"), "The name of the input port for the audio.").asString();
+    m_audioPort.open("/" + name + audioPortIn);
+
+    std::string emotionsPortOut = rf.check("emotions_output_port_name", yarp::os::Value("/emotions:o"), "The name of the output port for the emotions.").asString();
+
+    m_period = rf.check("period", yarp::os::Value(0.01), "The module period.").asDouble();
+
+    m_switchValue = rf.check("expression_counter_max_value", yarp::os::Value(2), "It controls the frequency with which the robot opens and close the mouth.").asInt();
+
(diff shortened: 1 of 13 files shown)


traversaro (approved):

S-Dafarra:
I just tested it on iCub3.

S-Dafarra:
@kouroshD if you agree I would merge

kouroshD:
Please let me review the PR before merging.

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
```diff
@@ -0,0 +1,181 @@
...
+        {
+            yError() << "Invalid frame length.";
+            return false;
+        }
+
+        if (m_isTalking)
+        {
+            m_switchCounter++;
+
+            if (m_switchCounter > m_switchValue)
```
kouroshD, line 142: So, here considering the values of `VAD_FREQUENCY` and  `VAD_SAMPLE_LENGTH`, while talking, every `20ms` we change the mouse values between `sur` and `nue`. Am I right @S-Dafarra ?

kouroshD, line 142: So, here considering the values of `VAD_FREQUENCY` and  `VAD_SAMPLE_LENGTH`, while talking, every `20ms` we change the mouse values between `sur` and `nue`. Am I right @S-Dafarra ?

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <yarp/os/LogStream.h>
+#include <functional>
+#include <cmath>
+
+#define VAD_FREQUENCY     8000 //Other frequencies do not seem to work well
+#define VAD_SAMPLE_LENGTH 80 //This seeemed to be the only value for which there was a usable output.
+
+bool FaceExpressionsRetargetingModule::configure(yarp::os::ResourceFinder &rf)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
```
kouroshD, line 20: Since, the process is single thread, is there a specific reason you added the mutex lock?

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
```diff
@@ -0,0 +1,181 @@
...
+bool FaceExpressionsRetargetingModule::close()
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    fvad_free(m_fvadObject);
+    m_fvadObject = nullptr;
+    m_audioPort.close();
+    m_emotionsOutputPort.close();
+    yInfo() << "Closing";
+
```
kouroshD, line 179: @S-Dafarra  Since the process is closing, isn't it cleaner to unlock the object to release the memory? However, the effect should be the same!

kouroshD (approved):
Thank you @S-Dafarra, for me is good to go!
I dropped only a few comments!

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
S-Dafarra: No those values are related to the processing of the input audio data. This ultimately depends on how often we receive this audio data, which depends on how the corresponding ``yarp`` device is configured. What we can configure is the ``expression_counter_max_value``. Here how it works:
- the operator speaks in the mic
- the audio is recorded in some buffers and sent to this module
- the module analyze the buffer and detects if someone is speaking -> ``m_isTalking > 0``

If someone is speaking, he might be speaking continuously. Hence, while ``m_isTalking > 0`` we want the mouth expression to toggle from "open" to "close" and vice-versa. There would be different way to implement the toggle. The easiest is through a counter. Every time ``m_isTalking > 0`` I increase a counter. When the counter is above a value, it toggles the face expression. Ultimatedly, the period of each iteration depends on how often we receive data from the microphone device.

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
S-Dafarra: If you refer to the ``lock``, that is scoped. So as soon as the ``close`` method returns, the ``mutex`` is automatically unlocked. The ``lock`` is necessary here because the ``close`` method might be called while ``updateModule`` is running on a different thread. Since I need to deallocate stuff, it is important I do this only when I am sure no other thread is accessing this data. Hence, it is necessary to lock the mutex.

modules/FaceExpressions_module/src/FaceExpressionsRetargeting.cpp
kouroshD: Thank you for the explanation, however it was clear from the code. The question was about the toggle time of the mouth while talking, which you are saying it is relevant to the audio `yarp` device configuration which is streaming the audio.

PR #58, opened 2021-05-27
Fix oculus xsens

kouroshD:
- fix the paramters for the joypad device
- remove imu option from `camCalibWithPose`
- remove the image orientation part since now the image is fixed wrt VR headset display
- remove the torso retargeting since it was only used for the image orientation in the case of xsens retargeting
- fix the head configuration in the case of  xsens retargeting in order not initialize it


S-Dafarra (approved):
I think you need to properly rebase on top of ``devel``. The modifications for the ``hde`` side should be already there.

kouroshD:
Moreover, I have tested only two demos with the new changes, Xsens+ virtualizer+ Oculus Walking, and Oculus+ walking. However, it should be OK for other demos, but it might be better to test them before as well.

kouroshD:
I rebased it on top of devel.

@S-Dafarra  @GiulioRomualdi  Should I merge, or we do a new round of tests before the merge?

kouroshD:
Following the f2f discussion, I'll do proceed with the merge.

PR #60, opened 2021-06-16
Add the possibility to use htc-vive in OculusModule

GiulioRomualdi:


modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -932,6 +983,42 @@ bool OculusModule::updateModule()
...
+                    yError() << "[OculusModule::updateModule] I will not start the walking. Please "
+                                "try to start again.";
+                    return true;
+                }
+
+                // get only the yaw axis
+                iDynTree::Rotation tempRot(getRotation(oculusInertial_T_oculusRoot));
+                double yaw = 0;
+                double dummy = 0;
+                tempRot.getRPY(dummy, dummy, yaw);
```
S-Dafarra, line 1012: Note that here you are assuming that the Z axis is parallel to gravity, but according to https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#reference-spaces it is the Y axis pointing upward

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -23,6 +23,29 @@
...
 #include <functional>
 
+Eigen::Ref<Eigen::Matrix<double, 3, 3, Eigen::RowMajor>>
+getRotation(yarp::sig::Matrix& m)
+{
+    assert(m.rows() == 4 && m.cols() = 4 && "getRotation expects a 4x4 matrix");
+    return Eigen::Map<Eigen::Matrix<double, 4, 4, Eigen::RowMajor>>(m.data()).topLeftCorner<3,3>();
+}
+
+auto getPosition(yarp::sig::Matrix& m)
```
S-Dafarra, line 33: I am not sure if this is ``c++14``.

modules/Oculus_module/include/OculusModule.hpp
```diff
@@ -79,6 +79,7 @@ class OculusModule : public yarp::os::RFModule
     bool m_useVirtualizer; /**< True if the virtualizer is used in the retargeting */
     bool m_useXsens; /**< True if the Xsens is used in the retargeting */
     bool m_useIFeel; /**< True if the iFeel is used in the retargeting */
+    bool m_useVive; /**< True if the iFeel is used in the retargeting */
```
S-Dafarra, line 82: Copy paste

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -598,8 +636,21 @@ bool OculusModule::getTransforms()
...
+                iDynTree::toEigen(m_oculusRoot_T_headOculus)
+                    = iDynTree::toEigen(m_oculusRoot_T_oculusInertial)
+                      * iDynTree::toEigen(m_oculusRoot_T_headOculus);
+            }
         }
     }
+
+    // m_oculusHeadsetPoseInertial is a 6d std::vector here I'm getting the first three elements
+    Eigen::Map<Eigen::Vector3d>(m_oculusHeadsetPoseInertial.data())
+        = getPosition(m_oculusRoot_T_headOculus);
```
S-Dafarra, line 652: What about the rotation?

modules/Oculus_module/src/OculusModule.cpp
GiulioRomualdi: `oculusInertial_T_oculusRoot` arrives from the transform server so it should contain a rotation matrix then we get only the yaw (idyntree definition). I think this does not depends on the name of the frame of openxr.

modules/Oculus_module/src/OculusModule.cpp
GiulioRomualdi: Ops. Here you have 
f96166e

modules/Oculus_module/include/OculusModule.hpp
GiulioRomualdi: 7ab3ab4

modules/Oculus_module/src/OculusModule.cpp
GiulioRomualdi: 😭  I'm not able to make the code working without auto. I should return the type of 
```
Eigen::Map<Eigen::Matrix<double, 4, 4, Eigen::RowMajor>>(m.data()).topRightCorner<3,1>()
```

S-Dafarra:
Closing this in favor of https://github.com/robotology/walking-teleoperation/pull/62

PR #66, opened 2021-07-15
Add TransformServer application

lrapetti:
As discussed in https://github.com/dic-iit/component_ifeel/pull/783, I am adding an application for running the `transformServer`

cc @S-Dafarra @kouroshD

PR #62, opened 2021-06-30
Allow head tracking for OpenXr headsets

S-Dafarra:

PR #69, opened 2021-09-16
Xprize demo June 2021 video updates

kouroshD:
This is a PR related to some changed that we collectively have intorduced for the Xprize June 2021 demo.
The changes are mainly related to the use of new headset and updates on the application file.
Moreover, I have added the yarpmanager subgroups for the June Demo.
Finally, I renamed the yarpmanager application files related to Xprize demos, to make them more clear.


modules/Oculus_module/src/HeadRetargeting.cpp
```diff
@@ -153,30 +187,55 @@ void HeadRetargeting::setDesiredHeadOrientation(
...
+
+    m_teleopFrame_R_headOculus
+        = iDynTree::Rotation::RotY(m_playerOrientation).inverse() //With OpenXr the Y is up
+            * m_oculusInertial_R_headOculus;
+
+    // notice here the following assumption is done:
+    // desiredNeckJoint(0) = neckPitch, the angle around X, with X pointing right
+    // desiredNeckJoint(1) = neckRoll, the angle around Z, with Z pointing backward
+    // desiredNeckJoint(2) = neckYaw, the angle around Y, with Y pointing up
+    inverseKinematicsXZY(
```
S-Dafarra, line 220:
```suggestion
    // The kinematic chain from the chest to the neck is composed of the pitch, roll, and yaw angles, in this order. 
    // The neck pitch axis is aligned with the x axis of the reference frame used by openxr, the roll with the z axis,
    // and the yaw with the y axis. Hence, we need to find the Euler angles corresponding to R_x  R_z * R_y.
    inverseKinematicsXZY(
```

S-Dafarra: Back in time, I did not comment this properly.

modules/Oculus_module/src/HeadRetargeting.cpp
```diff
@@ -64,6 +64,37 @@ void HeadRetargeting::inverseKinematics(const iDynTree::Rotation& chest_R_head,
     return;
 }
 
+void HeadRetargeting::inverseKinematicsXZY(const iDynTree::Rotation &chest_R_head,
```
S-Dafarra, line 67:
```suggestion
// This code was taken from https://www.geometrictools.com/Documentation/EulerAngles.pdf
// Section 2.2
void HeadRetargeting::inverseKinematicsXZY(const iDynTree::Rotation &chest_R_head,
```

S-Dafarra: I should have added a reference to how this implementation has been found.

S-Dafarra (approved):
Thanks @kouroshD, it looks good to me. Since you are adding the modifications introduced with https://github.com/robotology/walking-teleoperation/pull/62, I took the chance to add tow comments I had forgotten back in time

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -932,6 +994,43 @@ bool OculusModule::updateModule()
...
+
+                yarp::sig::Matrix openXrHeadInitialTransform = identitySE3();
+                if (!m_frameTransformInterface->getTransform(
+                        m_headFrameName, m_rootFrameName, openXrHeadInitialTransform))
+                {
+                    yError() << "[OculusModule::updateModule] Unable to evaluate the "
+                             << m_headFrameName << " to " << m_rootFrameName << "transformation";
+                    yError() << "[OculusModule::updateModule] I will not start the walking. Please "
+                                "try to start again.";
+                    return true;
```
kouroshD, line 1016: @S-Dafarra  here we are returning `true` because we would like to _wait_ until `getTransform(...)` returns true or there are some other reasons?

app/robots/iCubGenova09/oculusConfig.ini
```diff
@@ -45,9 +46,10 @@ rightCameraPort         /icub/cam/right
 [include TORSO_RETARGETING "torsoRetargeting.ini"]
 
 [OCULUS]
```
kouroshD, line 48: @S-Dafarra  Here instead of modifying the values of the `OCULUS` group, what we can do is to add `OPENXR` group and keep oculus group as it was and based on  `useOpenXr` we read one of the two options. In this case, if we want to do demo with iCub3 and oculus for any reason, we do not need to change these variables again.

kouroshD:
I will add this PR https://github.com/robotology/walking-teleoperation/pull/66 also to current PR, so that we have config files updated.

app/robots/iCubGenova09/oculusConfig.ini
S-Dafarra: That's agood point!

kouroshD:
> I will add this PR #66 also to current PR, so that we have config files updated.

Done.

app/robots/iCubGenova09/oculusConfig.ini
kouroshD: I have added this with the last commit. There are some other changes to the last file because of the clang formatting.
The two changes are in https://github.com/robotology/walking-teleoperation/pull/69/files#diff-18df315094c6bba8b4871f9c64c45a56e4ffb8e8c1e79fafedb34731d9dcef89 and https://github.com/robotology/walking-teleoperation/pull/69/files#diff-e53bdd7b4385b92cf93f7c48da41177f28a31e43275cc157c2f05c51c585e649R337-R339

kouroshD:
@S-Dafarra  Let me know about the two comments, and I will proceed with the metrging.

modules/Oculus_module/src/OculusModule.cpp
S-Dafarra: That ``if`` is in the ``updateModule`` method and it is reached when starting the state machine. If we have that error, we return true to avoid continuing (thus the internal state machine will not change state), and we avoid closing the module (as it happens when returning false).

app/robots/iCubGenova09/oculusConfig.ini
```diff
@@ -51,6 +52,14 @@ right_hand_frame_name           roculus
...
 
+[OPENXR]
+root_frame_name                 openxr_origin
+left_hand_frame_name            openxr_left_hand
+right_hand_frame_name           openxr_right_hand
+head_frame_name                 openxr_head
+oculusOrientationPort           /oculusOrientation:i
+oculusPositionPort              /oculusPosition:i
+
 move_icub_using_joypad        1
```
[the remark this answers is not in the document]
kouroshD: done.

modules/Oculus_module/src/OculusModule.cpp
kouroshD: Yes, so I understood correctly your intention that it was intentionally set to true to avoid closing the module and wait for the transformation :)

kouroshD:
@S-Dafarra I will continue with the merging if there is no other comments.

S-Dafarra (approved):
Go for me!

kouroshD:
> Go for me!

Done, thanks. It will be great if you can test the `devel` branch during these demos before merging it to `master` branch.

PR #70, opened 2021-09-21
add rpc server to the oculus module to control finite state machine of the module

kouroshD:
Problem: 
Especially by using the VIVE, it is often tricky to distinguish the left from the right hand. It would be better and safer to start the teleoperation pipeline from a terminal command
Solution:
An RPC command is added to the oculus module.

### DIFF
--- a/modules/Oculus_module/include/OculusModule.hpp
+++ b/modules/Oculus_module/include/OculusModule.hpp
@@ -273,21 +273,6 @@ public:
     bool runningModule();
 
     /**
-     * Respond to a message from the RPC port.
-     * @param command is the received message.
-     * The following message has to be a bottle with the following structure:
-     * 1. ("help");
-     * 2. ("prepare").
-     * 3. ("run")
-     * @param reply is the response of the server.
-     * 1. OK in case of success;
-     * 2. Not OK in case of failure.
-     * 3. list of available commands in case of "help" command
-     * @return true in case of success and false otherwise.
-     */
-    bool respond(const yarp::os::Bottle& command, yarp::os::Bottle& reply) final;
-
-    /**
      * to get the Oculus FSM in string format
      * @param OculusFSM to get the string value
      * @return string associated with the inpute sate
--- a/modules/Oculus_module/include/OculusModule.hpp
+++ b/modules/Oculus_module/include/OculusModule.hpp
@@ -22,6 +22,7 @@
 #include <yarp/os/BufferedPort.h>
 #include <yarp/os/RFModule.h>
 #include <yarp/os/RpcClient.h>
+#include <yarp/os/RpcServer.h>
 #include <yarp/sig/Vector.h>
 
 #include <FingersRetargeting.hpp>
@@ -129,11 +130,8 @@ private:
     yarp::os::RpcClient
         m_rpcVirtualizerClient; /**< Rpc client used for sending command to the virtualizer */
 
-    /** Port used to retrieve the human whole body joint pose. */
-    yarp::os::BufferedPort<yarp::os::Bottle> m_wholeBodyHumanJointsPort;
+    yarp::os::RpcServer m_rpcOculusServer; /**< RPC port to control Oculus FSM. */
 
-    /** Port used to retrieve the human whole body joint pose. */
-    yarp::os::BufferedPort<yarp::sig::Vector> m_wholeBodyHumanSmoothedJointsPort;
 
     double m_robotYaw; /**< Yaw angle of the robot base */
 
@@ -244,6 +242,28 @@ public:
      * @return true in case of success and false otherwise.
      */
     bool close() final;
+
+     /**
+     * Respond to a message from the RPC port.
+     * @param command is the received message.
+     * The following message has to be a bottle with the following structure:
+     * 1. ("help");
+     * 2. ("prepare").
+     * 3. ("run")
+     * @param reply is the response of the server.
+     * 1. OK in case of success;
+     * 2. Not OK in case of failure.
+     * 3. list of available commands in case of "help" command
+     * @return true in case of success and false otherwise.
+     */
+    bool respond(const yarp::os::Bottle& command, yarp::os::Bottle& reply) final;
+
+    /**
+     * to get the Oculus FSM in string format
+     * @param OculusFSM to get the string value
+     * @return string associated with the inpute sate
+     */
+    std::string getStringFromOculusState(const OculusFSM state);
 };
 
 inline std::string getTimeDateMatExtension()
--- a/modules/Oculus_module/include/OculusModule.hpp
+++ b/modules/Oculus_module/include/OculusModule.hpp
@@ -13,6 +13,7 @@
 // std
 #include <ctime>
 #include <memory>
+#include <mutex>
 
 // YARP
 #include <yarp/dev/IFrameTransform.h>
@@ -29,6 +30,8 @@
 #include <HandRetargeting.hpp>
 #include <HeadRetargeting.hpp>
 
+#include <thrifts/TeleoperationCommands.h>
+
 #ifdef ENABLE_LOGGER
 #include <matlogger2/matlogger2.h>
 #include <matlogger2/utils/mat_appender.h>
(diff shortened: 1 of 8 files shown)


kouroshD:
@S-Dafarra  @GiulioRomualdi  if you can test this feature for the demo, it would be great. I could build it without any problem, but I could not test it.

S-Dafarra:
I quickly checked it. Here a couple of comments:
1. the ``RPC`` answer is asynchronous, hence you need to use some ``mutex`` around.
2. I would suggest using separate methods dealing with each command.
3. Instead of using the ``respond`` method, you can specify a thrift with all the commands you need (http://www.yarp.it/git-master/thrift_tutorial_simple.html). This has the following advantages:
    - for each command it automatically generates a virtual method that you can implement, thus automatically addressing point 2 above;
    - it handles automatically the ``help`` command;
    - it handles automatically more complex inputs, like vectors
    - if you write some documentation, this is automatically reported when typing ``help myCommand``
    - it is more scalable; it is easier to add/change/remove commands

You can find some example of usage of the thrifts in the ``walking-controller``:
- the thrift file is https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/thrifts/WalkingCommands.thrift
- this is the command to launch in the CMakeLists https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/CMakeLists.txt#L37-L38 this will automatically create some files that you need to add to the executable https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/CMakeLists.txt#L42.
- In the module then, you include the autogenerated header https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/include/WalkingControllers/WalkingModule/Module.h#L53. This has the same name of the thrift file. 
- you inherit from the autogenerated class (that has the same name of the thrift file): https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/include/WalkingControllers/WalkingModule/Module.h#L61
- You implement the methods related to the commands: https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/include/WalkingControllers/WalkingModule/Module.h#L258-L288
- You still have the RPC port https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/include/WalkingControllers/WalkingModule/Module.h#L120
- You use this commands to "attach" the methods you implemented to the RPC port https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/src/Module.cpp#L170-L176
- You close the port normally: https://github.com/robotology/walking-controllers/blob/75429cbbf444c92496beb863667cac0c50eb542e/src/WalkingModule/src/Module.cpp#L356

kouroshD:
@S-Dafarra  I have updated the PR with the requested changes. Please have a look on it.
Thanks.

modules/Oculus_module/include/OculusModule.hpp
```diff
@@ -244,6 +247,52 @@ class OculusModule : public yarp::os::RFModule
...
+     * 1. ("help");
+     * 2. ("prepare").
+     * 3. ("run")
+     * @param reply is the response of the server.
+     * 1. OK in case of success;
+     * 2. Not OK in case of failure.
+     * 3. list of available commands in case of "help" command
+     * @return true in case of success and false otherwise.
+     */
+    bool respond(const yarp::os::Bottle& command, yarp::os::Bottle& reply) final;
```
S-Dafarra, line 288: I think this is not necessary anymore

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -553,6 +566,266 @@ bool OculusModule::close()
...
+        //                            + ", while it should be in state: "
+        //                            + getStringFromOculusState(OculusFSM::InPreparation));
+        //        }
+    } else
+    {
+        reply.addString("Entered command is incorrect, Enter `help` to know available commands");
+    }
+
+    return true;
+}
```
S-Dafarra, line 713: I think this should not be necessary anymore

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -553,6 +566,266 @@ bool OculusModule::close()
...
+        yWarning() << "The teleoperation is in state: " + getStringFromOculusState(m_state)
+                          + ", while it should be in state: "
+                          + getStringFromOculusState(OculusFSM::InPreparation)
+                   << "in order to run the teleoperation.";
+        yWarning() << "No update on the teleoperation state, try again.";
+        return false;
+    }
+
+    return true;
+}
```
S-Dafarra, line 746: This is not doing anything

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -967,19 +1240,12 @@ bool OculusModule::updateModule()
...
         {
-            // TODO add a visual feedback for the user
-            if (m_moveRobot)
-            {
-                cmd.addString("prepareRobot");
-                m_rpcWalkingClient.write(cmd, outcome);
-            }
-            m_state = OculusFSM::InPreparation;
-            yInfo() << "[OculusModule::updateModule] prepare the robot";
+            std::lock_guard<std::mutex> guard(m_mutex);
```
S-Dafarra, line 1245: The lock should not just be here. For example, in the lines above you are accessing to the ``m_state`` variable, that can be edited at anytime by the RPC calls. The scope of the lock should be on the entire ``updateModule``.

S-Dafarra (changes requested):
I think the lock of the mutex is missing in other places, like the ``close`` method.

modules/Oculus_module/include/OculusModule.hpp
kouroshD: Yes, I should remove it, it is a left over, thanks.

kouroshD:
@S-Dafarra  Thank you for the comments, I have updated the PR with your comments.

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -518,7 +531,10 @@ bool OculusModule::configure(yarp::os::ResourceFinder& rf)
         yInfo() << "[OculusModule::configure] Cameras have been reset.";
     }
 
-    m_state = OculusFSM::Configured;
+    {
+        std::lock_guard<std::mutex> guard(m_mutex);
```
S-Dafarra, line 535: I would lock the entire method. ``m_state`` is not the only variable that could be affected by concurrent read/write

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -553,6 +569,149 @@ bool OculusModule::close()
...
+        yWarning() << "No update on the teleoperation state, try again.";
+        return false;
+    }
+
+    return this->runningModule();
+}
+
+bool OculusModule::preparingModule()
+{
+    std::lock_guard<std::mutex> guard(m_mutex);
```
S-Dafarra, line 636: This is a deadlock. If ``preparingModule`` is called by other methods, then it should not attempt to lock the mutex if the caller is also locking it

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -553,6 +569,149 @@ bool OculusModule::close()
...
+    }
+    m_state = OculusFSM::InPreparation;
+    yInfo() << "[OculusModule::preparingModule] prepare the robot";
+
+    return true;
+}
+
+bool OculusModule::runningModule()
+{
+    std::lock_guard<std::mutex> guard(m_mutex);
```
S-Dafarra, line 651: This is also a deadlock

S-Dafarra (changes requested):

kouroshD:
@S-Dafarra  I have updated the PR with your suggestions. Moreover, I have rebased the code. Let me know if you have other comments as well. Thanks.

kouroshD:
We have tested PR added to this branch :  https://github.com/robotology/walking-teleoperation/tree/modify_dcmWalkingRetargeting_rpc_Oculus
It was working as expected.
@S-Dafarra  If you can approve the PR, I will proceed with the merge.

S-Dafarra (approved):

S-Dafarra:
Just one comment for future reference. When using the gloves now, we could avoid connecting to the ``JoypadControlServer``.

PR #72, opened 2021-10-07
Add Haptic Glove Module

kouroshD:
- [ ] I still have to add README file to the module
- [x] I have to fix the configuration files
- [x] I have to check the possible changes outside the scope of this PR on other modules and remove those changes if are unnecessary
- [x] After the latest changes, I have to run the code with the robot.


kouroshD:
The PR is ready to be reviewed, except the points mentioned in the first comment.

kouroshD:
I have tested many times during these days the code, and I applied bug fixes as well as updates on the configuration file. All the features are implemented. The only remaining point is the documentation, that I may do it later. In any case, the PR is ready to be reviewed.

S-Dafarra (commented):
Some preliminary feedback

traversaro:
Given the size of the PR, if @S-Dafarra is already reviewing it I would remove myself from the reviewers. If there is any specific point on which you need my input @kouroshD feel free to point it out, thanks!

modules/HapticGlove_module/src/HapticGloveModule.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if (m_useRightHand)
+    {
+        if (!m_rightHand->close())
+        {
+            yError() << m_logPrefix << "unable to close right hand teleoperation.";
+            return false;
+        }
+    }
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(100)); // wait for 100ms.
```
S-Dafarra, line 135: Why is this needed?

modules/HapticGlove_module/src/LinearRegression.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+using namespace HapticGlove;
+
+LinearRegression::LinearRegression()
+{
+}
+
+LinearRegression::~LinearRegression()
+{
+}
```
S-Dafarra, line 11: In general, if you need to specify a custom destructor, you would need to specify also the custom copy/move constructor and copy/move assignment. This is the so called [``rul of five``](https://en.cppreference.com/w/cpp/language/rule_of_three). In general, if you can avoid specifying a custom destructor, it is better.

Sometimes the destructor is needed when using this object inside a smart pointer. I am not sure if this is the case.

modules/HapticGlove_module/src/KalmanFilter.cpp
```diff
@@ -0,0 +1,146 @@
...
+     * J= 1/2 [(x-x_bar) M^(-1)(x-x_bar) + (z-Hx) R^(-1)(z-Hx)]
+     */
+
+    m_z = z;
+
+    m_P = (m_M.inverse() + m_Ht_Rinv_H).inverse();
+    m_K = m_P * m_Ht_Rinv;
+    m_x_hat = m_x_bar + m_K * (m_z - m_H * m_x_bar);
+    m_x_bar = m_Phi * m_x_hat + m_Gamma * m_w_bar;
+    m_M = m_Phi * m_P * m_Phi.transpose() + m_Gamma * m_Q * m_Gamma.transpose();
```
S-Dafarra, line 95: Note, if these computations are called during the control loop, Eigen might need to create temporary objects that may slow down the execution (because they need dynamic memory allocation). See https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html

I am not sure this is the case anyways

modules/HapticGlove_module/src/LinearRegression.cpp
```diff
@@ -0,0 +1,61 @@
...
+                                    const CtrlHelper::Eigen_Mat& ouputData,
+                                    CtrlHelper::Eigen_Mat& tetha)
+{
+
+    // TODEL
+    std::cout << "[LinearRegression::LearnOneShot()]" << std::endl;
+
+    CtrlHelper::Eigen_Mat xT_x = (inputData.transpose() * inputData);
+    std::cout << "xT_x: " << xT_x.rows() << xT_x.cols() << std::endl;
+    std::cout << "xT_x.determinant(): " << xT_x.determinant() << std::endl;
```
S-Dafarra, line 29: The computation of the determinant might be expensive. Not sure if it is worth for a simple debug message

modules/HapticGlove_module/src/MotorEstimation.cpp
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+#include <MotorEstimation.hpp>
+#include <iostream>
+using namespace HapticGlove;
+
+Estimator::Estimator(const double& dt, const Eigen::MatrixXd& R, const Eigen::MatrixXd& Q)
+{
+    // hard coded model
+    // reference issue: https://github.com/ami-iit/element_retargeting-from-human/issues/145
```
S-Dafarra, line 16: Nitpick, this code is public, but this issue is not public. I think it is ok to keep the link, but maybe you can add some details

modules/HapticGlove_module/src/RobotInterface.cpp
```diff
@@ -0,0 +1,1216 @@
...
+        return false;
+    }
+
+    m_noAnalogSensor = analogList.size();
+
+    std::vector<bool> jointFeedbackIsAnalog(m_noAllJoints, false);
+    for (size_t i = 0; i < m_allJointNames.size(); i++)
+    {
+        if (find(std::begin(analogList), std::end(analogList), m_allJointNames[i])
+            != std::end(analogList))
```
S-Dafarra, line 76: Suggestion. The ``find`` moves linearly across the ``analogList`` vector (hence the complexity is linear) once for every joint. You can save some time if the ``find`` complexity was constant. This is possible using some more complex data structure like [``std::unordered_set``](https://en.cppreference.com/w/cpp/container/unordered_set/find). Basically, the data is saved using some hashing mechanism that allows finding an entry in constant time. This may also come useful if you have to scan the vector multiple times. It is really not a big deal here since you are in the configuration phase, but it might be helpful in other cases.

modules/HapticGlove_module/src/RobotInterface.cpp
```diff
@@ -0,0 +1,1216 @@
...
+    {
+        yError() << m_logPrefix << "cannot obtain IPWMControl interface";
+        return false;
+    }
+
+    if (!m_robotDevice.view(m_pidInterface) || !m_pidInterface)
+    {
+        yError() << m_logPrefix << "cannot obtain IPidControl interface";
+        return false;
+    }
```
S-Dafarra, line 338: I would suggest moving this to a dedicated method

modules/HapticGlove_module/src/RobotMotorsEstimation.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        estimatedValues.resize(m_numOfMotors, 0.0);
+    }
+
+    if (estimatedAccelerations.size() != m_numOfMotors)
+    {
+        estimatedValues.resize(m_numOfMotors, 0.0);
+    }
+
+    if (P.rows() != m_numOfMotors && P.cols() == std::pow(m_n, 2))
```
S-Dafarra, line 136: Nitpick, ``std::pow`` is usually more expensive than simply ``m_n * m_n``. See https://stackoverflow.com/questions/2940367/what-is-more-efficient-using-pow-to-square-or-just-multiply-it-with-itself

S-Dafarra (approved):
I mostly skimmed through the code. Most of the comments are simply suggestions

modules/HapticGlove_module/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+  include/Logger.hpp
+  include/Teleoperation.hpp
+  include/ControlHelper.hpp
+  )
+
+# add include directories to the build.
+include_directories(
+  ${CMAKE_CURRENT_SOURCE_DIR}/include
+#  ${SenseGlove_INCLUDE_DIR}
+)
```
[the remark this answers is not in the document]
kouroshD: Without `include_directories` the compiler will throw an error  `No such file or directory` since it cannot find the file.

modules/HapticGlove_module/include/RobotMotorsEstimation.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * perform an estimation step assuming stochastic steady state system
+     * @param z new measurement vector
+     */
+    bool estimateNextSteadyState(const std::vector<double>& z);
+
+    /**
+     * get the estimation results
+     * @param estimatedValue expected value
```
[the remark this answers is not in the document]
kouroshD: The angle of joint/axis. Fixed.

modules/HapticGlove_module/src/GloveControlHelper.cpp
```diff
@@ -0,0 +1,335 @@
...
+        m_desiredVibrotactileValues[i]
+            = (int)std::round(std::max(0.0, std::min(desiredValue[i], 100.0)));
+    }
+
+    return m_pImp->setFingertipVibrotactileValues(m_desiredVibrotactileValues);
+}
+bool GloveControlHelper::stopPalmVibrotactileFeedback()
+{
+    return m_pImp->setPalmVibrotactileValue(
+        124); // to turn off set `124`, for more details check the documentation.
```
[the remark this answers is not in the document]
kouroshD: here is the different types of commands we can set to the glove plam buzz motor: https://github.com/robotology/wearables/blob/ff6bd84a98f3fc01dc44e09af24475c2c72ce1ae/devices/HapticGlove/include/SenseGloveHelper.hpp#L38-L70
As you can see, the order of those values have not specific meaning. Moreover, through wearable we can set a value to wearable port. Here a guide is provided what are the permissible values.

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        wearableActuatorCommand.value = values[i];
+        wearableActuatorCommand.info.name = m_wearablePrefix
+                                            + wearable::actuator::IHaptic::getPrefix() + fingerName
+                                            + "::ForceFeedback";
+        wearableActuatorCommand.info.type = wearable::msg::ActuatorType::HAPTIC;
+        wearableActuatorCommand.info.status = wearable::msg::ActuatorStatus::OK;
+        wearableActuatorCommand.duration = 0;
+
+        m_iWearActuatorPort.write(true);
```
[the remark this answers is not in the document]
kouroshD: fixed.

modules/HapticGlove_module/src/HapticGloveModule.cpp
kouroshD: Removed from here. In any case, the there is sleep in the close method of teleoperation class right after sending `stopHapticFeedback()` in order to be sure the stop haptic command is sent before closing the module, otherwise the glove may continue to provide the haptic feedback according to the last sent command.

app/robots/iCubGazeboV2_5/leftFingersHapticRetargetingParams.ini
```diff
@@ -0,0 +1,156 @@
+###############
+### ROBOT
+###############
+remote_control_boards   ("left_arm")
+
+remote_sensor_boards    "left_hand"
+
+axis_list             (  "l_hand_fingers", "l_thumb_oppose", "l_thumb_proximal", "l_thumb_distal", "l_index_proximal", "l_index_distal" , "l_middle_proximal" , "l_middle_distal" , "l_pinky" )
+
+all_axis_list         (  "l_hand_fingers", "l_thumb_oppose", "l_thumb_proximal", "l_thumb_distal", "l_index_proximal", "l_index_distal" , "l_middle_proximal" , "l_middle_distal" , "l_pinky" )
```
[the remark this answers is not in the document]
kouroshD: The [IAxisInfo](https://github.com/robotology/yarp/blob/master/src/libYARP_dev/src/yarp/dev/IAxisInfo.h) provides some functionalities like to get the axis name given the iD of the joint/axis. But, I am not sure if we can get the name of all the axis names associated to a control board. Moreovr, the orders of the all_axis_list is important. Finally, some time, the axis may not show at the SW level any problem, but we see clearly they are not working fine even if not in hw fault. So, I would  keep the `axis_list` as the only variable in this config file which allows the user to control the axis of interest.

kouroshD:
With this https://github.com/robotology/walking-teleoperation/pull/72/commits/fc89789c4fabb663a023cae91fd3c22e512bb5b0, I have modified the `writeStrict` option of the `remotecontrolboard` from `true` to `false`.

kouroshD:
another thing is that in the `RemoteControlBoard` the default carrier is set to `udp` as in https://github.com/robotology/yarp/blob/ce6f008c86ba58a652b04b6bffdc2b46e9b157e2/src/devices/RemoteControlBoard/RemoteControlBoard.cpp#L247-L250 and https://github.com/robotology/yarp/blob/ce6f008c86ba58a652b04b6bffdc2b46e9b157e2/src/devices/RemoteControlBoard/RemoteControlBoard.cpp#L307. And I remember, I could run the `HapticGloveModule` some months ago from the operator side as well, however with some difficulties.  In any case, now we are running it the robot head.

modules/HapticGlove_module/include/RobotInterface.hpp
```diff
@@ -0,0 +1,518 @@
+/**
+ * @file RobotInterface.hpp
+ * @authors Kourosh Darvish <kourosh.darvish@iit.it>
+ * @copyright 2020 iCub Facility - Istituto Italiano di Tecnologia
+ *            Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ * @date 2020
+ */
+
+#ifndef ROBOT_INTERFACE_HPP
```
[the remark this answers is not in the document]
kouroshD: Let's keep as it is. In future for other projects, if there is the name clash, we can modify it always.

modules/HapticGlove_module/src/LinearRegression.cpp
kouroshD: Thanks for the suggestion. I set all the destructors to `default`.

app/robots/iCubGazeboV2_5/leftFingersHapticRetargetingParams.ini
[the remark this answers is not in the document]
S-Dafarra: Ok clear!

modules/HapticGlove_module/include/RobotController.hpp
```diff
@@ -0,0 +1,302 @@
...
+     * @return true in case of success and false otherwise.
+     */
+    bool setAxisReferences(const Eigen::VectorXd& axisReferences);
+
+    /**
+     * Set the fingers joint reference value
+     * @param fingersReference the reference value for the finger joints to follow
+     * @return true in case of success and false otherwise.
+     */
+    bool setJointReferences(const std::vector<double>& fingersReferences);
```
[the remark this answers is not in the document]
S-Dafarra: Clear, thanks!

modules/HapticGlove_module/src/GloveControlHelper.cpp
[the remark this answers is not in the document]
S-Dafarra: My suggestion was to add a method in https://github.com/robotology/wearables/blob/ff6bd84a98f3fc01dc44e09af24475c2c72ce1ae/devices/HapticGlove/include/SenseGloveHelper.hpp to disable the palm feedback. In this way, if you want to use other gloves that use different numbers, you don't have to change this line. Do you see what I mean?

modules/HapticGlove_module/src/HapticGloveModule.cpp
S-Dafarra: I see. The important thing is to write why there is that delay so that it is understandable in the future.

modules/HapticGlove_module/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  include/Retargeting.hpp
+  include/LinearRegression.hpp
+  include/GloveWearable.hpp
+  include/Logger.hpp
+  include/Teleoperation.hpp
+  include/ControlHelper.hpp
+  )
+
+# add include directories to the build.
+include_directories(
```
S-Dafarra, line 45: Rather than using ``include_directories``, it is better to include ``target_include_directories``. See https://stackoverflow.com/questions/31969547/what-is-the-difference-between-include-directories-and-target-include-directorie

modules/HapticGlove_module/src/GloveControlHelper.cpp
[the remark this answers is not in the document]
kouroshD: Yes, I know what you mean and it sounds reasonable. The problem is related to wearable interface which does not allow to have such interface. Through Wearables, we can only set a value for a haptic type name and stream it to the dedicated port.
What instead we can do, to avoid having that magic number (which has already the explanation), I copy https://github.com/robotology/wearables/blob/ff6bd84a98f3fc01dc44e09af24475c2c72ce1ae/devices/HapticGlove/include/SenseGloveHelper.hpp#L38-L70 to walking-teleoperation project, and set by enum value and doing the casting.
Let me know @S-Dafarra  if is sounds like a good compromise?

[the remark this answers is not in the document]
S-Dafarra: I am still not convinced. That enum contains a series of magic numbers and tags (e.g. ``Cue_Game_Over``) that make sense only for the senseglove.

I would have expected the driver to reinterpret a generic input signal (for example 0 -> no feedback, 124 -> max feedback, or a more straightforward set of tags) into something that the senseglove would understand. 

In any case, from what I understood, in order to clean this part, we would also need another PR on the wearable side. Since this is already complex on its own, I would suggest spawning a specific issue on this and move on.

modules/HapticGlove_module/src/KalmanFilter.cpp
kouroshD: Yes, this is called in the control loop. I have added `noalias()` method before the `=` operator, to take care of that when having a matrix product or inversion. I have made this update all around code.
Thanks for pointing this out.

S-Dafarra: You can use what defined in http://eigen.tuxfamily.org/index.php?title=FAQ#Where_in_my_program_are_temporary_objects_created.3F to understand if some temporary objects are created. If there are, the suggestion would be to use variables to store the intermediate steps.

modules/HapticGlove_module/src/RobotInterface.cpp
kouroshD: I see your point. Thank you for the explanation. As you mentioned here, we are in the configuration time and it should be OK. Moreover, the order of elements of the elements of the `analogList` is important in our case, and as the elements are not sorted in any particular order in [`unordered_set`](https://en.cppreference.com/w/cpp/container/unordered_set) it may cause problem. As you can see in https://github.com/robotology/walking-teleoperation/blob/fc89789c4fabb663a023cae91fd3c22e512bb5b0/modules/HapticGlove_module/src/RobotInterface.cpp#L125 , we are considering the index of the `analogList` for future data association.

modules/HapticGlove_module/CMakeLists.txt
kouroshD: agree, fixed.

kouroshD:
@S-Dafarra  let me know if the responses are OK, if it needs further declarations/modifications.

S-Dafarra:
I guess you can also remove that line. By default oit should be ``off``

modules/HapticGlove_module/src/GloveControlHelper.cpp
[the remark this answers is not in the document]
kouroshD: > I guess you can also remove that line. By default it should be `off`

It is not. Look at here: 

https://github.com/robotology/yarp/blob/165a46c6e034fc27b90cc0bd80b35419679b22c3/src/devices/RemoteControlBoard/RemoteControlBoard.h#L96-L97

kouroshD:
> I guess you can also remove that line. By default it should be `off`

It is not. Look at here: 

https://github.com/robotology/yarp/blob/165a46c6e034fc27b90cc0bd80b35419679b22c3/src/devices/RemoteControlBoard/RemoteControlBoard.h#L96-L97

S-Dafarra:
> > I guess you can also remove that line. By default it should be `off`
> 
> It is not. Look at here:
> 
> https://github.com/robotology/yarp/blob/165a46c6e034fc27b90cc0bd80b35419679b22c3/src/devices/RemoteControlBoard/RemoteControlBoard.h#L96-L97

Ah nice catch. It seems that it enables it when writing to multiple joints. I guess this is to make sure that the commands sent to all the joints is coherent in time :thinking: .

modules/HapticGlove_module/src/KalmanFilter.cpp
kouroshD: @S-Dafarra  @traversaro   I added the assertion check as in http://eigen.tuxfamily.org/index.php?title=FAQ#Where_in_my_program_are_temporary_objects_created.3F . However, I did the check with or WITHOUT `.noalias()` ; both the times the results were similar. No message showing the heap memory allocation. However, for the matrix product it should be case. Am I doing sth wrong?
Here is the commit  https://github.com/robotology/walking-teleoperation/pull/72/commits/ab05a12f3c8096c9810aff5e2f6fc903e855b5d1 
 
I added permanently the option to CMake and made it as false by default.

I had some compilation error while adding the command as mentioned in  https://github.com/stack-of-tasks/tsid/issues/7 and benefit from the implementation in https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/CMakeLists.txt and https://github.com/stack-of-tasks/tsid/blob/devel/include/tsid/math/fwd.hpp

S-Dafarra: To be honest I never tried it myself. From my side, this is not blocking at all.

IN any case, have you tried compiling in debug? Or intentionally add some other matrix in the product?

kouroshD: No, I tried in Release mode, not in `Debug` mode. 

> Or intentionally add some other matrix in the product?

No, I just remove `.noalias()` from the member variables.

kouroshD: OK, I tried in debug mode and returns the assertion results. The assertion fails in the following line:
```
    m_P = (m_M.inverse() + m_Ht_Rinv_H).inverse();
```
which is expected.

The following line the assertion passes:
```
    m_K = m_P * m_Ht_Rinv;
```
While again, the following line the assertion does not pass, which would expect it to pass:
```
    m_x_hat = m_x_bar + m_K * (m_z - m_H * m_x_bar);
```
and here is the back-trace results:

```
HapticGloveModule: /usr/include/eigen3/Eigen/src/Core/util/Memory.h:143: void Eigen::internal::check_that_malloc_is_allowed(): Assertion `is_malloc_allowed() && "heap allocation is forbidden (EIGEN_RUNTIME_NO_MALLOC is defined and g_is_malloc_allowed is false)"' failed.

Thread 1 "HapticGloveModu" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff75d1859 in __GI_abort () at abort.c:79
#2  0x00007ffff75d1729 in __assert_fail_base (fmt=0x7ffff7767588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x555555627490 "is_malloc_allowed() && \"heap allocation is forbidden (EIGEN_RUNTIME_NO_MALLOC is defined and g_is_malloc_allowed is false)\"", 
    file=0x555555627458 "/usr/include/eigen3/Eigen/src/Core/util/Memory.h", line=143, function=<optimized out>) at assert.c:92
#3  0x00007ffff75e2f36 in __GI___assert_fail (
    assertion=0x555555627490 "is_malloc_allowed() && \"heap allocation is forbidden (EIGEN_RUNTIME_NO_MALLOC is defined and g_is_malloc_allowed is false)\"", 
    file=0x555555627458 "/usr/include/eigen3/Eigen/src/Core/util/Memory.h", line=143, function=0x555555627420 "void Eigen::internal::check_that_malloc_is_allowed()")
    at assert.c:101
#4  0x000055555557512b in Eigen::internal::check_that_malloc_is_allowed () at /usr/include/eigen3/Eigen/src/Core/util/Memory.h:143
#5  0x0000555555575144 in Eigen::internal::aligned_malloc (size=24) at /usr/include/eigen3/Eigen/src/Core/util/Memory.h:155
#6  0x0000555555580105 in Eigen::internal::conditional_aligned_malloc<true> (size=24) at /usr/include/eigen3/Eigen/src/Core/util/Memory.h:214
#7  0x000055555557e00d in Eigen::internal::conditional_aligned_new_auto<double, true> (size=3) at /usr/include/eigen3/Eigen/src/Core/util/Memory.h:374
#8  0x000055555557a398 in Eigen::DenseStorage<double, -1, -1, -1, 1>::resize (this=0x55555587e968, size=3, rows=3, cols=1)
    at /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:408
#9  0x0000555555577356 in Eigen::PlainObjectBase<Eigen::Matrix<double, -1, -1, 1, -1, -1> >::resize (this=0x55555587e968, rows=3, cols=1)
    at /usr/include/eigen3/Eigen/src/Core/PlainObjectBase.h:293
#10 0x0000555555589050 in Eigen::internal::resize_if_allowed<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, double, double> (
    dst=..., src=...) at /usr/include/eigen3/Eigen/src/Core/AssignEvaluator.h:720
#11 0x0000555555585412 in Eigen::internal::call_dense_assignment_loop<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::internal::assign_op<double, double> > (dst=..., src=..., func=...) at /usr/include/eigen3/Eigen/src/Core/AssignEvaluator.h:734
#12 0x00005555555837fa in Eigen::internal::Assignment<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::internal::assign_op<double, double>, Eigen::internal::Dense2Dense, void>::run (dst=..., src=..., func=...) at /usr/include/eigen3/Eigen/src/Core/AssignEvaluator.h:879
#13 0x00005555555823f9 in Eigen::internal::call_assignment_no_alias<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::internal::assign_op<double, double> > (dst=..., src=..., func=...) at /usr/include/eigen3/Eigen/src/Core/AssignEvaluator.h:836
#14 0x00005555555d036b in Eigen::internal::assignment_from_xpr_op_product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::CwiseBinaryOp<Eigen::internal::scalar_difference_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, 0> const>, 0>, Eigen::internal::assign_op<double, double>, Eigen::internal::add_assign_op<double, double> >::run<Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::CwiseBinaryOp<Eigen::internal::scalar_difference_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, 0> const>, 0> const>, Eigen::internal::assign_op<double, double> > (dst=..., src=...) at /usr/include/eigen3/Eigen/src/Core/ProductEvaluators.h:223
#15 0x00005555555cf813 in Eigen::internal::call_assignment_no_alias<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::CwiseBinaryOp<Eigen::internal::scalar_difference_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, 0> const>, 0> const>, Eigen::internal::assign_op<double, double> > (dst=..., src=..., func=...) at /usr/include/eigen3/Eigen/src/Core/AssignEvaluator.h:836
#16 0x00005555555ce608 in Eigen::NoAlias<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::MatrixBase>::operator=<Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::CwiseBinaryOp<Eigen::internal::scalar_difference_op<double, double>, Eigen::Matrix<double, -1, -1, 1, -1, -1> const, Eigen::Product<Eigen::Matrix<double, -1, -1, 1, -1, -1>, Eigen::Matrix<double, -1, -1, 1, -1, -1>, 0> const>, 0> const> > (this=0x7fffffffc570, other=...) at /usr/include/eigen3/Eigen/src/Core/NoAlias.h:42
#17 0x00005555555cd0f3 in HapticGlove::KalmanFilter::estimateNextState (this=0x55555587e820, z=...)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/KalmanFilter.cpp:95
#18 0x00005555555de666 in HapticGlove::Estimator::estimateNextState (this=0x5555558435a0, z=...)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/MotorEstimation.cpp:82
#19 0x00005555555e0fba in HapticGlove::Estimators::estimateNextState (this=0x555555755720, z=std::vector of length 9, capacity 9 = {...})
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/RobotMotorsEstimation.cpp:97
#20 0x0000555555574372 in HapticGlove::RobotController::estimateNextStates (this=0x5555556e65f0)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/RobotController.cpp:511
#21 0x000055555561ee76 in HapticGlove::Teleoperation::getFeedbacks (this=0x555555712bc0)
--Type <RET> for more, q to quit, c to continue without paging--
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/Teleoperation.cpp:184
#22 0x000055555561f3ac in HapticGlove::Teleoperation::run (this=0x555555712bc0)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/Teleoperation.cpp:239
#23 0x00005555555c6b7e in HapticGloveModule::updateModule (this=0x7fffffffc910)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/HapticGloveModule.cpp:146
#24 0x00007ffff7c02dc4 in yarp::os::RFModule::runModule() () from /home/kourosh/icub_ws/robotology-superbuild/build/install/lib/libYARP_os.so.3
#25 0x00007ffff7c0328e in yarp::os::RFModule::runModule(yarp::os::ResourceFinder&) ()
   from /home/kourosh/icub_ws/robotology-superbuild/build/install/lib/libYARP_os.so.3
#26 0x000055555556dff8 in main (argc=1, argv=0x7fffffffcb48)
    at /home/kourosh/icub_ws/robotology-superbuild/src/walking-teleoperation/modules/HapticGlove_module/src/main.cpp:40
(gdb) 
```
here it does not  create temporary objects:
`    m_x_bar = m_Phi * m_x_hat + m_Gamma * m_w_bar;
`
while in `    m_M = m_Phi * m_P * m_Phi.transpose() + m_Gamma * m_Q * m_Gamma.transpose();` it does make new objects. I tried to introduce new variable for `m_Phi.transpose()` and `m_Gamma.transpose()` but for those parts also new objects are dynamically allocated. 
I stop the analysis at this point. I will keep the `noalias()` as it should not hurt  as it has been carefully used.

kouroshD: So, finally I am going to keep the changes introduced in this commit https://github.com/robotology/walking-teleoperation/pull/72/commits/e94eb10ad4cbdf3b47ea37914ef41cbab0faa2cb , as it will not change anything if the option is not set `ON` by the user and if it is not compiled in `Debug` mode. I will not do further optimization at this stage, since it is not limiting. @S-Dafarra if you agree I resolve this conversation with the outputs of the investigation that has been done.

traversaro:
> > > I guess you can also remove that line. By default it should be `off`
> > 
> > 
> > It is not. Look at here:
> > https://github.com/robotology/yarp/blob/165a46c6e034fc27b90cc0bd80b35419679b22c3/src/devices/RemoteControlBoard/RemoteControlBoard.h#L96-L97
> 
> Ah nice catch. It seems that it enables it when writing to multiple joints. I guess this is to make sure that the commands sent to all the joints is coherent in time 🤔 .

Yes, writeStrict is necessary for all the commands that require multiple messages for different joints, otherwise tipically the last one would be the only one considered.

modules/HapticGlove_module/src/KalmanFilter.cpp
S-Dafarra: Yes ok for me. 
Just notice that the fact that the error disappears when compiling in Release does not mean that it is fixed. When compiling in Release all the asserts (including the one above) disappear. So, the issue is still there, simply you "silenced" the message.

Approximately, the only operations that do not trigger a memory allocation are the sum, the transpose and a simple matrix product. The sum of two matrix products does (see https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html). 
This is why 
```
    m_x_hat = m_x_bar + m_K * (m_z - m_H * m_x_bar);
```
produces a temporary object. You need to store the result of ``m_H * m_x_bar`` somewhere before applying the difference. The same also for ``m_K * (m_z - m_H * m_x_bar)``.

Basically, it becomes a tradeoff between the readability of the code and its execution speed.

S-Dafarra:
So then, maybe it is worth leaving it with the default behavior :thinking:

kouroshD:
> Yes, writeStrict is necessary for all the commands that require multiple messages for different joints, otherwise tipically the last one would be the only one considered.

@traversaro  What do you mean by `requiring multiple messages for different joints`? Can you give an example?

modules/HapticGlove_module/src/GloveControlHelper.cpp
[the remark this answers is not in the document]
kouroshD: With this commit https://github.com/robotology/walking-teleoperation/pull/72/commits/8d336ef917ebb993c42b0bcd05db10d9d30db316, I have added the enum class for the senseGlove thumper commands. And not the stop commands use the enum value to send stopping value.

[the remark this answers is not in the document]
kouroshD: related issue https://github.com/robotology/walking-teleoperation/issues/86

Closing this conversation. It can followed in the issue.

modules/HapticGlove_module/src/KalmanFilter.cpp
kouroshD: > Just notice that the fact that the error disappears when compiling in Release does not mean that it is fixed. When compiling in Release all the asserts (including the one above) disappear. So, the issue is still there, simply you "silenced" the message.


Yes, I know!


> Approximately, the only operations that do not trigger a memory allocation are the sum, the transpose and a simple matrix product. The sum of two matrix products does (see https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html).
This is why
```
    m_x_hat = m_x_bar + m_K * (m_z - m_H * m_x_bar);
```
produces a temporary object. You need to store the result of m_H * m_x_bar somewhere before applying the difference. The same also for m_K * (m_z - m_H * m_x_bar).

I tried last night, it needs many intermediate variables to add. I am not in the favour of it as it is not a blocking point at this stage, as you mentioned, it will be readable! Probably, even this one `m_K * (m_z - m_H * m_x_bar).` needs an intermediate member variable, since we have two matrix product.

For example I was trying to do it for `m_Phi * m_P * m_Phi.transpose() + m_Gamma * m_Q * m_Gamma.transpose();` we need one variable for each transpose, and then another one for single matrix product  and then another one for each operand of `+` and ... . I really prefer in this way to be readable at this stage.

kouroshD:
> > > Yes, writeStrict is necessary for all the commands that require multiple messages for different joints, otherwise tipically the last one would be the only one considered.
> > 
> > 
> > @traversaro What do you mean by `requiring multiple messages for different joints`? Can you give an example?
> 
> Let's say that for some reason you are sending velocity reference joint-by-joint with: https://github.com/robotology/yarp/blob/165a46c6e034fc27b90cc0bd80b35419679b22c3/src/devices/RemoteControlBoard/RemoteControlBoard.cpp#L1914 . In that case, if you do not have writeStrict, only the last command will be considered.

As I see in that snippet, it sends the joint `i` reference value with `false` write strict option. I am not seeing any part that is discarding other commands! booh! I can put it back in our code in default option, what is your suggestion? When I tried yesterday with `off` option, I did not see any strange behavior. In any case, in our code, we are always send vector of joint values not a single joint!

kouroshD:
@S-Dafarra  @traversaro  If you do not have other comments or suggestion, I will merge the PR. Thanks.

traversaro:
> @S-Dafarra @traversaro If you do not have other comments or suggestion, I will merge the PR. Thanks. I will keep the history of PR as it is.

Sure on my side, see https://github.com/robotology/walking-teleoperation/pull/72#issuecomment-1008733201 !

S-Dafarra:
> @S-Dafarra @traversaro If you do not have other comments or suggestion, I will merge the PR. Thanks. I will keep the history of PR as it is.

GO also for me!

modules/HapticGlove_module/src/GloveWearable.cpp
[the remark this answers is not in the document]
kouroshD: @S-Dafarra  @traversaro  @diegoferigo Following this comment, which we set the `writeStrict` to `false`, the effect is that, it is not correctly sending the haptic feedback for all the fingers. It only sends the thumb haptic feedback and the rest is ignored (to be more precise `HapticGlove::haptic::l_thumb_finger::ForceFeedback` which is the first one to set). Any suggestion? should I revert it back to `true` in the next PR, or there might be some other solutions!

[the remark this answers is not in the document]
kouroshD: I have tested that with the real robot. Actually in simulation is the same, it is clear in the simulation video I putted here https://github.com/robotology/walking-teleoperation/pull/72#issuecomment-1014657236 , in the right-bottom terminal.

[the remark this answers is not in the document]
S-Dafarra: I guess this is because every message you write in ``m_iWearActuatorPort`` have different meaning every time, and it is not just the same data sent over and over with a different value.

I guess this is one of the cases where ``forceStrict`` was indeed necessary. 

I think it is ok to open a PR even for just a line modification, it will be just easier to review :grin:

PR #73, opened 2021-11-18
Use transform server

S-Dafarra:


S-Dafarra:
It needs to be tested yet

S-Dafarra:
While testing, I noticed https://github.com/robotology/walking-teleoperation/issues/74

S-Dafarra:
I tested it. Here how it looks


https://user-images.githubusercontent.com/18591940/142446962-8822734a-b9b8-4731-b7f4-a472f8694b12.mp4

modules/Virtualizer_module/src/VirtualizerModule.cpp
```diff
@@ -375,6 +438,11 @@ bool VirtualizerModule::updateModule()
         return false;
     }
 
+    //get player height
+    float playerHeightInCm;
+    playerHeightInCm = m_cvirtDeviceID->GetPlayerHeight(); //It is relative to the initial height set ResetPlayerHeight, positive upward
+    double playerHeighInm = playerHeightInCm / 100.0;
```
kouroshD, line 444:
```suggestion
    double playerHeighInM = playerHeightInCm / 100.0;
```

modules/Virtualizer_module/src/VirtualizerModule.cpp
```diff
@@ -352,6 +412,9 @@ bool VirtualizerModule::close()
     m_encodersInterface = nullptr;
     m_controlModeInterface = nullptr;
 
+    m_tfDriver.close();
+    m_tfPublisher = nullptr;
```
kouroshD, line 416:
```suggestion
    m_tfPublisher = nullptr;
    m_tfDriver.close();
```

kouroshD: I am not sure also if `m_useTf` is set to false, reaching this point while closing causes segfault or will close normally.

modules/Virtualizer_module/src/VirtualizerModule.cpp
```diff
@@ -467,6 +535,22 @@ bool VirtualizerModule::updateModule()
...
+        Eigen::Matrix3d rotation = Eigen::AngleAxisd(-playerYaw, Eigen::Vector3d::UnitZ()).toRotationMatrix(); //We consider the Z axis pointing upward, hence we invert the sign since the angle from the virtualizer is positive clockwise
+
+        for (size_t i = 0; i < 3; ++i)
+        {
+            for (size_t j = 0; j < 3; ++j)
+            {
+                m_tfMatrix(i,j) = rotation(i,j);
+            }
+        }
+        m_tfMatrix(2,3) = playerHeighInm; //Set the z position equal to the height
```
kouroshD, line 549:
```suggestion
        m_tfMatrix(2,3) = playerHeighInM; //Set the z position equal to the height
```

kouroshD (commented):
Thank you @S-Dafarra
There are a few minor comments that are added.

One more is, to add the `--use_transform_server true` to the xml files and possibly adding the dependency to the transform server when this option is enabled. This probably can be done in https://github.com/robotology/walking-teleoperation/pull/75

modules/Virtualizer_module/src/VirtualizerModule.cpp
S-Dafarra: The close is supposed to be safe (i.e. if nothing was opened, it should simply return). The order of the expressions does not really matter.

PR #75, opened 2021-11-22
App modifications

S-Dafarra:
This PR extracts the application modifications from https://github.com/robotology/walking-teleoperation/tree/modify_dcmWalkingRetargeting_rpc_Oculus


S-Dafarra:
The audio still need to be checked

isorrentino (approved):

S-Dafarra:
This is now blocked by #76

app/scripts/Xprize-VisualRetargeting.xml
```diff
@@ -19,26 +19,27 @@
...
-    <name>camCalibWithPose</name>
+    <name>camCalib</name>
     <parameters>--context cameraCalibration --from icubEyes.ini --group CAMERA_CALIBRATION_RIGHT --name /icub/camcalib/right</parameters>
     <node>icub-console</node>
   </module>
 
   <module>
     <name>yarpdev</name>
-    <parameters>--device JoypadControlServer --use_separate_ports 1 --period 10 --name /joypadDevice/Oculus --subdevice openxrheadset --stick_as_axis true --gui_elements 1 --GUI_0::width 0.15 --GUI_0::height 0.15 --GUI_0::x 0.2 --GUI_0::y -0.2 --GUI_0::z -0.5</parameters>
+    <parameters>--device JoypadControlServer --use_separate_ports 1 --period 10 --name /joypadDevice/Oculus --subdevice openxrheadset --stick_as_axis true --gui_elements 1 --GUI_0::width 0.15 --GUI_0::height 0.15 --GUI_0::x 0.15 --GUI_0::y -0.2 --GUI_0::z -0.5</parameters>
```
kouroshD, line 35: Probably, here we can use `--from dcmWalkingRetargeting.ini ` instead of this long line of list of parameters.
Should be checked, if all the variables are correctly set then in the `.ini` file.

kouroshD (commented):
Thank you @S-Dafarra  for the PR. 
Two small points to be considered before merging.

kouroshD:
The new file names related to the `JoypadControlServer` device parameters is called `dcmWalkingRetargeting.ini`. Probably, it is copy-pasting problem, since the configuration file name is misleading. Probably, sth like `JoypadControlServerParameters.ini` can be better representative of those parameters.

app/scripts/Xprize-VisualRetargeting.xml
```diff
@@ -95,7 +70,7 @@
   <connection>
     <from>/idyntree-yarp-visualizer/image</from>
     <to>/joypadDevice/Oculus/gui_0</to>
-    <protocol>mjpeg</protocol>
+    <protocol>fast_tcp</protocol>
```
kouroshD, line 73: Why we are changing it? If not wrong, `mjpeg` doing some compression of the data before buffering it to the network. Normally, for images, `mjpeg`is recommended to be used.

app/scripts/Xprize-KynDynRetargeting.xml
```diff
@@ -17,36 +17,37 @@
     <node>icub-head</node>
     <parameters>--from dcm_walking_iFeel_joint_retargeting.ini</parameters>
   </module>
-  
+
 
   <module>
     <name>VirtualizerModule</name>
+    <parameters>--use_transform_server true</parameters>
```
kouroshD, line 24:
```suggestion
    <parameters>--use_transform_server true</parameters>
    <dependencies>
          <port timeout="15.0">/transformServer/transforms:o</port>
     </dependencies>
```

kouroshD: @S-Dafarra  I've added this dependency to the transform server port, since with the proposed option we have dependency on it. If it is OK for you, please consider committing it.

S-Dafarra: Good idea! I just removed an extra space.

kouroshD:
I've rebased the branch after #76  and #73 .

PR #76, opened 2021-11-23
Allow the Oculus retargeting to start also without joypads

S-Dafarra:


modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -216,6 +196,43 @@ bool OculusModule::configureJoypad(const yarp::os::Searchable& config)
...
+    {
+        if (!m_useSenseGlove || !m_useVirtualizer)
+        {
+            yError() << "[OculusModule::configureJoypad] Unable to open the polydriver.";
+            return false;
+        }
+        else
+        {
+            yWarning() << "[OculusModule::configureJoypad] Unable to open the polydriver. "
+                          "Continuing anyway since we using the virtualizer and the gloves.";
```
isorrentino, line 232: >  "Continuing anyway since we using the virtualizer and the gloves.";

 "Continuing anyway since we are using the virtualizer and the gloves.";

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -1030,10 +1066,14 @@ bool OculusModule::updateModule()
         }
 
         // check if it is time to prepare or start walking
-        float buttonMapping;
+        float buttonMapping = 0.0;
```
kouroshD, line 1069: Probably, it is safer to initialize it with `-1.0`, since we have later a check if the value is bigger than zero.

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -1122,10 +1162,13 @@ bool OculusModule::updateModule()
     } else if (m_state == OculusFSM::Configured)
     {
         // check if it is time to prepare or start walking
-        float buttonMapping;
+        float buttonMapping = 0.0;
```
kouroshD, line 1165: Probably, it is safer to initialize it with `-1.0`, since we have later a check if the value is bigger than zero.

modules/Oculus_module/src/OculusModule.cpp
```diff
@@ -1146,9 +1189,12 @@ bool OculusModule::updateModule()
             }
         }
 
-        float buttonMapping;
-        // start walking (X button)
-        m_joypadControllerInterface->getButton(m_startWalkingIndex, buttonMapping);
+        float buttonMapping = 0.0;
```
kouroshD, line 1192: similarly to previous comments, here!

kouroshD:
Since we have two similar configuration parameters (i.e., `skip_joypad` and `move_icub_using_joypad`), for the sake of completeness and future reference let's describe them here:
The joypad is used for the following two set of functionalities:
- To continuously move the robot, or provide references to the robot:
  -  The reference to the walking controller for the robot base motion (when we are not using virtualizer)
  - The reference velocities to the robot finger velocities (when not using SenseGlove)
 - To change the walking-teleoperation state machine and prepare and start the teleoperation

For these functionalities, we have the following relevant configuration parameters:
- `move_icub_using_joypad` --> to Not use Virtualizer
- `useSenseGlove` --> to user SenseGlove
and now:
- `skip_joypad` --> To totally avoid using joypad for any functionality.

So, if we are using both the `Virtualizer` and `SenseGlove`, we can skip the joypad (since we can use also rpc command to change the state machine https://github.com/robotology/walking-teleoperation/pull/70). Otherwise (if for one of those two functionalities, we are not using `SenseGlove` or `Virtualizer`), we cannot skip jpypad.

@S-Dafarra  Maybe at this point instead of `move_icub_using_joypad`, we can change it to `useVirtualizer`? But for this, we have to update all the laptop files, and the all the configuration files.

kouroshD (approved):
I have added a few minor comments, please have a look on them. If you agree, you may add the changes. Let me know, then we can follow the merge of this PR.

S-Dafarra:
> @S-Dafarra Maybe at this point instead of `move_icub_using_joypad`, we can change it to `useVirtualizer`? But for this, we have to update all the laptop files, and the all the configuration files.

This makes sense, but I think it is outside the scope of this PR.

kouroshD:
I've rebased the branch after #73 , merging.

PR #84, opened 2022-01-11
Add WeightRetargeting to Xprize Semifinal yarpmanager Application File

RiccardoGrieco:


app/scripts/Xprize-UserInterfacesRetargeting.xml
```diff
@@ -123,5 +119,11 @@
     <to>/joypadDevice/Oculus/rpc</to>
     <protocol>fast_tcp</protocol>
   </connection>
+  
+  <connection>
+    <from>/WeightRetargeting/output:o</from>
+    <to>/iFeelSuit/WearableActuatorsCommand/input:i</to>
+    <protocol>udp</protocol>
```
kouroshD, line 126: Depending on where we run the module, at the moment we may not have the udp option for the carrier type. Now, as @S-Dafarra  was mentioning before, we cannot have udp for the double-network configuration. In our case, I think the module will run on robot side, so udp is not a reliable solution. We may use fast_tcp, what do you say @RiccardoGrieco ?

app/scripts/Xprize-UserInterfacesRetargeting.xml
```diff
@@ -97,18 +97,14 @@
...
-    <name>iFeelSuitSkinActuatorsTestApplication</name>
+    <name>WeightRetargetingModule</name>
+    <parameters>--from WeightRetargetingElbows.ini</parameters>
     <environment>YARP_FORWARD_LOG_ENABLE=1</environment>
-    <description>Run iFeelSuit sking application</description>
+    <description>Run weight retargeting application</description>
     <node>localhost</node>
-    <tag>iFeelSuit-SkinActuatorsTestApplication</tag>
-    <dependencies>
-      <port timeout="11.0">/iFeelSuit/WearableActuatorsCommand/input:i</port>
```
kouroshD, line 108: Maybe we can keep the dependencies that are valid for this module, what do you say @RiccardoGrieco ?!

kouroshD (approved):

kouroshD:
thank you @RiccardoGrieco . I made a small comment, when you are done, let me know, I will merge the PR. thanks.

app/scripts/Xprize-UserInterfacesRetargeting.xml
S-Dafarra: To be honest, I am not sure anymore about this. I remember that the audio did not work when using ``udp``. Maybe it was an isolated case, or maybe it got fixed in the meantime. In any case, I agree that ``fast_tcp`` is safer

RiccardoGrieco:
Thanks @kouroshD for the heads-up!

If you agree I would wait for the following issues to be closed to merge the PR, since for instance the configuration files might change.

- https://github.com/ami-iit/element_weight-retargeting/issues/14
- https://github.com/ami-iit/element_weight-retargeting/issues/17

app/scripts/Xprize-UserInterfacesRetargeting.xml
RiccardoGrieco: Done in fdd66be

app/scripts/Xprize-UserInterfacesRetargeting.xml
RiccardoGrieco: I agree! Done in c88f246.

RiccardoGrieco:
>@RiccardoGrieco I would say, we can try to have the latest commands that allows us to run those codes. Later, we can easily update the yarpmanager application. If still they are the same, please let's merge this PR. I would like to merge the upstream branch to devel.

@kouroshD changed the name of the config file to a generic one, so we can keep using it while correcting/deciding the internal parameters. Should have used it earlier 😄.

The PR can be merged.

PR #90, opened 2022-02-11
Fix Utils compilation on windows

S-Dafarra:


kouroshD:
I am wondering why this happens when you were compiling. CI does not throw any error, neither on my system.
Probably, we can rename the PR title to `Utils` compilation @S-Dafarra ?

kouroshD (approved):

S-Dafarra:
> I am wondering why this happens when you were compiling. CI does not throw any error, neither on my system. Probably, we can rename the PR title to `Utils` compilation @S-Dafarra ?

Sure! I don't know why this did not occur on CI. I wondered if it could be related to a ``yarp`` version, although the CI is using ``master``. In any case, it is worth to add that include since you we are using ``std::time_t`` in that file.

kouroshD:
thanks merged.

PR #92, opened 2022-03-03
Eyes tracking hysteresis

S-Dafarra:
Improved the gaze retargeting by adding a series of hysteresis and quantizations

PR #94, opened 2022-04-13
Virtualizer velocity deadzone

S-Dafarra:
This is on top of https://github.com/robotology/walking-teleoperation/pull/92

Keeping this in draft until that is merged


kouroshD:
The code looks fine, however, the strategy implemented, and the parameters set here seems to me very specific to xprize semifinal competition setup.
Are we sure we want to merge it?

kouroshD (approved):

S-Dafarra:
> The code looks fine, however, the strategy implemented, and the parameters set here seems to me very specific to xprize semifinal competition setup.

The value of the parameters is indeed Xprize related. I would like to merge the code we used for the semifinals in the master branches. The value of the parameters can change in future PRs.

Regarding the strategy, why do you think it is very specific to Xprize?

PR #96, opened 2022-05-06
Fix openxr singularity

S-Dafarra:
When starting the ``OculusRetargetinModule``, the mechanism that consider the initial head orientation is subject to singularities when the headset starts with a rotation of 180deg around gravity. This fixes it

This PR is on top of #94


S-Dafarra:
Friendly ping @GiulioRomualdi

S-Dafarra:
This has been merged with https://github.com/robotology/walking-teleoperation/pull/103

PR #97, opened 2022-05-19
Fix deprecations of yarp methods

S-Dafarra:

PR #100, opened 2022-06-08
Texture retargeting

kouroshD:
Probably we need to rebase on top of `devel` due to https://github.com/robotology/walking-teleoperation/pull/97.

As I mentioned before, I do not set up my systems now, so I cannot do it now.
@S-Dafarra  you can simply rebase it on top of `devel`, and build to check. You should also fix the deprecation problems as in https://github.com/robotology/walking-teleoperation/pull/97 later. If you cannot, I can do it in these days, but not today for sure.
**Avoid doing any squashing, I would like to keep the commit history clean.**

### DIFF
--- a/app/robots/iCubErzelli02/hapticGloveConfig.ini
+++ b/app/robots/iCubErzelli02/hapticGloveConfig.ini
@@ -19,6 +19,11 @@ getHumanMotionRange     1
 steadyStateCounterThreshold 5
 # the threshold to check the error between the desired and feedback values for the steady state check [rad]
 steadyStateThreshold        0.05
+# threshold in which an skin data is considered updated [no units]
+tactileWorkingThreshold     0.0001
+# threshold in which an skin data is considered updated [no units]
+tactileUpdateThreshold     0.0001
+
 
 # include fingers parameters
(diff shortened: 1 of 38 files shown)


S-Dafarra:
> @S-Dafarra you can simply rebase it on top of `devel`, and build to check.

Done. I also changed the base branch to ``master``



> You should also fix the deprecation problems as in #97

Done in https://github.com/robotology/walking-teleoperation/pull/100/commits/2fa747650467d912c175c09a6bd095e0182e0c51

S-Dafarra (approved):
Given the circumstances, we can simply merge this PR as is once the CI passes

kouroshD:
> Given the circumstances, we can simply merge this PR as is once the CI passes

If you want to have a look to the PR and code (however, it is a bit large), I invite you to do so. If there are suggestions, I can fix it in the following days. In any case, I have been carefully updated the code to be clean and with clear necessary explanation and documentation.

I am also available later on, if you have some doubts or feedbacks!

kouroshD:
Should I continue with the merge @S-Dafarra ?!

S-Dafarra:
I quickly skimmed through the code and seems good to me! I also appreciated the documentation in the configuration files and in the code! Thanks a million, @kouroshD. Merging!

PR #125, opened 2023-03-07
Added files for ergoCub and modified SRanipal to control the eyes tilt only

S-Dafarra:


GiulioRomualdi (approved):
We may think to avoid installing the application of the yarp manager and do something similar to robots-configuration/extra

S-Dafarra:
> We may think to avoid installing the application of the yarp manager and do something similar to robots-configuration/extra

Good point. I have opened https://github.com/robotology/walking-teleoperation/issues/126 to discuss about this.

In the meantime, I would proceed with this PR and eventually fix all the applications later.

PR #135, opened 2023-10-30
updated name of the YARP port from HDE

davidegorbani:
After having merged https://github.com/robotology/human-dynamics-estimation/pull/367 it will be necessary to merge also this PR in order to use the `yarpmanager` applications since the name of the YARP ports exposed by the devices in HDE are changed.

### DIFF
--- a/app/scripts/Faire-UserRetargeting-iFeel.xml
+++ b/app/scripts/Faire-UserRetargeting-iFeel.xml
@@ -57,8 +57,8 @@
     <node>icub-console</node>
     <tag>RobotStateProvider</tag>
     <dependencies>
-      <port timeout="15.0">/iCub/RobotStateWrapper/state:o</port>
-      <port timeout="15.0">/HDE/WearableTargetsWrapper/state:o</port>
+      <port timeout="15.0">/iCub/RobotStateServer/state:o</port>
+      <port timeout="15.0">/HDE/WearableTargetsServer/state:o</port>
     </dependencies>
   </module>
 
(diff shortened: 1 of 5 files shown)


lrapetti:
@S-Dafarra @GiulioRomualdi I don't know if there is, or if it would make sense to have, a minimum required version for the dependency, such that avoid installing non-compatible configuration files. In that case we might add
```
find_package(HumanDynamicsEstimation 3.0.0 REQUIRED)
```
after realeasing `human-dynamics-estimation v3.0.0` (https://github.com/robotology/human-dynamics-estimation/issues/373)

lrapetti:
cc @mebbaid

S-Dafarra:
> @S-Dafarra @GiulioRomualdi I don't know if there is, or if it would make sense to have, a minimum required version for the dependency, such that avoid installing non-compatible configuration files. In that case we might add
> 
> ```
> find_package(HumanDynamicsEstimation 3.0.0 REQUIRED).
> ```
> 
> after realeasing `human-dynamics-estimation v3.0.0` ([robotology/human-dynamics-estimation#373](https://github.com/robotology/human-dynamics-estimation/issues/373))

Good point, the dependency is specified in https://github.com/robotology/walking-teleoperation/blob/da230ab1177698cb2dc9032916b585a33deacc30/cmake/WalkingTeleoperationFindDependencies.cmake#L147-L148

lrapetti:
> I missed that, so we can:
> 
> * add the dependency on v3.0.0. (even if its not out yet)
> * wait to merge it until release is done
> * not add the dependency, or add it later

As discussed at today meeting we will avoid adding the dependency.

davidegorbani:
I rebased the branch and reverted the modification to the Xprise files.

S-Dafarra (approved):
Thanks @davidegorbani! Merging

Issue #136, opened 2023-11-07
Fix the transmission strategy of the haptic feedback


S-Dafarra:
Right now, both the force and the vibrotactile feedback are sent in a single command per finger using write strict. See
https://github.com/robotology/walking-teleoperation/blob/a3b51def52b3c420d123331a03f924019e0e57f6/modules/HapticGlove_module/src/GloveWearable.cpp#L338-L354 and https://github.com/robotology/walking-teleoperation/blob/a3b51def52b3c420d123331a03f924019e0e57f6/modules/HapticGlove_module/src/GloveWearable.cpp#L368-L385

This is terribly inefficient, requires a huge bandwidth, and it is blocking the module in case of connection issues.

cc @EhsanRanjbari

EhsanRanjbari:
Is there a way to measure and print the bandwidth?  @S-Dafarra

S-Dafarra:
We were using ``iftop``. I am not sure if there is a way to get it programmatically.

EhsanRanjbari:
I have some ideas in order to decrease the bandwidth usage by the haptic glove. First, let's start by seeing how I understood the issue so far: 

![hapticCommand](https://github.com/robotology/walking-teleoperation/assets/75568997/7fa617ec-7622-411e-9df2-f60ad34ffbac)

So, HapticGloveModule writes the haptic command to the buffer port, and on the other side, the `HapticGloveDevice` reads the haptic command from the buffer port. 

`HapticGloveModule`, writes the haptic command using the `write()` method with the `forceStrict = true` option which is equivalent to the method `writeStrict()` ([check here](https://yarp.it/latest/classyarp_1_1os_1_1BufferedPort.html#af24f1371b310763572aa7f22459fd71d)). 

https://github.com/robotology/walking-teleoperation/blob/a3b51def52b3c420d123331a03f924019e0e57f6/modules/HapticGlove_module/src/GloveWearable.cpp#L352

So, from the figure above, the total bandwidth usage by the haptic glove is `bandwidth_1 + bandwidth_2.`

My idea is first,
1. Decrease the number of commands sent in total by NOT sending the zero-value commands: **_Decrease bandwidth_1_**
[...]

PR #139, opened 2024-01-30
Fix Haptic glove transmission - HapticGloveModule

EhsanRanjbari:
See https://github.com/robotology/walking-teleoperation/issues/136#issuecomment-1916788820

### DIFF
--- a/modules/HapticGlove_module/include/GloveControlHelper.hpp
+++ b/modules/HapticGlove_module/include/GloveControlHelper.hpp
@@ -40,6 +40,7 @@ class HapticGlove::GloveControlHelper
 
     bool m_isRightHand; /**< true if the glove is the right hand*/
 
+    std::vector<int> m_desiredHapticValues; /**< Desired haptic feedback values; range: [0, 100]. */
     std::vector<int> m_desiredForceValues; /**< Desired force feedback values; range: [0, 100]. */
     std::vector<int>
         m_desiredVibrotactileValues; /**< Desired vibrotactile feedback values; range: [0, 100] */
@@ -94,6 +95,12 @@ public:
     bool getHandJointAngles(std::vector<double>& jointAngles);
 
     /**
+     * Set the desired haptic feedback reference to the user fingertip
+     * @return true/false in case of success/failure
+     */
+    bool setFingertipHapticFeedbackReferences();
+
+    /**
      * Set the desired force feedback reference to the user fingertip
      * @param desiredValue desired force feedback values
      * @return true/false in case of success/failure
@@ -114,6 +121,12 @@ public:
      * @return true / false in case of success/failure
      */
     bool setPalmVibrotactileFeedbackReference(const int& desiredValue);
+
+    /**
+     * stop the haptics feedback (force feedback, fingertip vibrotactile feedback) to the user
+     * @return  true/false in case of success/failure
+     */
+    bool stopHapticsFeedback();
 
     /**
      * stop the haptic feedback (force feedback, fingertip vibrotactile feedback, palm vibrotactile
--- a/modules/HapticGlove_module/include/GloveControlHelper.hpp
+++ b/modules/HapticGlove_module/include/GloveControlHelper.hpp
@@ -126,7 +126,7 @@ public:
      * stop the haptics feedback (force feedback, fingertip vibrotactile feedback) to the user
      * @return  true/false in case of success/failure
      */
-    bool stopHapticsFeedback();
+    bool stopAllHapticFeedback();
 
     /**
      * stop the haptic feedback (force feedback, fingertip vibrotactile feedback, palm vibrotactile
--- a/modules/HapticGlove_module/src/GloveControlHelper.cpp
+++ b/modules/HapticGlove_module/src/GloveControlHelper.cpp
@@ -179,7 +179,7 @@ bool GloveControlHelper::setFingertipForceFeedbackReferences(
                               / m_maxForceFeedback);
     }
 
-    return m_pImp->setFingertipForceFeedbackValues(m_desiredForceValues);
+    return true;
 }
 
 bool GloveControlHelper::setFingertipVibrotactileFeedbackReferences(
@@ -200,22 +200,12 @@ bool GloveControlHelper::setFingertipVibrotactileFeedbackReferences(
             = (int)std::round(std::max(0.0, std::min(desiredValue[i], 100.0)));
     }
 
-    return m_pImp->setFingertipVibrotactileValues(m_desiredVibrotactileValues);
+    return true;
 }
 
 bool GloveControlHelper::setFingertipHapticFeedbackReferences()
 {
-
-    for (size_t i = 0; i < m_numForceFeedback; i++)
-    {
-        m_desiredHapticValues[i] = m_desiredForceValues[i];
-    }
-    for (size_t i = 0; i < m_numVibrotactileFeedback; i++)
-    {
-        m_desiredHapticValues[i + m_numForceFeedback] = m_desiredVibrotactileValues[i];
-    }
-
-    return m_pImp->setFingertipHapticFeedbackValues(m_desiredHapticValues);
+    return m_pImp->setFingertipHapticFeedbackValues(m_desiredForceValues, m_desiredVibrotactileValues);
 }
 
 bool GloveControlHelper::stopPalmVibrotactileFeedback()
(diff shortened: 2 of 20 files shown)


EhsanRanjbari:
> @EhsanRanjbari could you please rebase on top of master?

@S-Dafarra, it should be fixed now.

modules/HapticGlove_module/include/GloveControlHelper.hpp
```diff
@@ -115,6 +122,12 @@ class HapticGlove::GloveControlHelper
      */
     bool setPalmVibrotactileFeedbackReference(const int& desiredValue);
 
+    /**
+     * stop the haptics feedback (force feedback, fingertip vibrotactile feedback) to the user
+     * @return  true/false in case of success/failure
+     */
+    bool stopHapticsFeedback();
```
S-Dafarra, line 129: Nitpick. This does not seem grammatically correct. What about ``stopAllHapticFeedback``?

modules/HapticGlove_module/src/GloveControlHelper.cpp
```diff
@@ -201,6 +202,22 @@ bool GloveControlHelper::setFingertipVibrotactileFeedbackReferences(
...
+bool GloveControlHelper::setFingertipHapticFeedbackReferences()
+{
+
+    for (size_t i = 0; i < m_numForceFeedback; i++)
+    {
+        m_desiredHapticValues[i] = m_desiredForceValues[i];
+    }
+    for (size_t i = 0; i < m_numVibrotactileFeedback; i++)
+    {
+        m_desiredHapticValues[i + m_numForceFeedback] = m_desiredVibrotactileValues[i];
```
S-Dafarra, line 215: Would it be possible to use separate vectors?

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -329,61 +330,40 @@ bool GloveWearableImpl::getFingertipPoseValues(Eigen::MatrixXd& values)
...
-        wearableActuatorCommand.value = values[i];
-        wearableActuatorCommand.info.name = m_wearablePrefix
-                                            + wearable::actuator::IHaptic::getPrefix() + fingerName
-                                            + "::ForceFeedback";
-        wearableActuatorCommand.info.type = wearable::msg::ActuatorType::HAPTIC;
-        wearableActuatorCommand.duration = 0;
-
-        m_iWearActuatorPort.write(true); // writeStrict option for wearable haptic device should be
-                                         // set to true to avoid the data loss for all actuators
-    }
```
S-Dafarra, line 354: If this now does not do anything, is the method still useful?

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -329,61 +330,40 @@ bool GloveWearableImpl::getFingertipPoseValues(Eigen::MatrixXd& values)
...
     return true;
 }
 
 bool GloveWearableImpl::setFingertipVibrotactileValues(const std::vector<int>& values)
 {
-    if (values.size() != m_numVibrotactileFeedback)
+    return true;
```
S-Dafarra, line 338: Same here, if you are simply returning true, it means that this method should not be used anymore. Is there a reason to keep it?

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -329,61 +330,40 @@ bool GloveWearableImpl::getFingertipPoseValues(Eigen::MatrixXd& values)
...
+                    "haptic feedback size.";
     }
+    wearable::msg::WearableActuatorCommand& wearableActuatorCommand
+        = m_iWearActuatorPort.prepare();
+    wearableActuatorCommand.forceValue.resize(m_numForceFeedback, 0);
+    wearableActuatorCommand.vibroTactileValue.resize(m_numVibrotactileFeedback, 0);
+    wearableActuatorCommand.duration.resize(m_numForceFeedback + m_numVibrotactileFeedback, 0);
 
-    for (size_t i = 0; i < values.size(); i++)
+    for (size_t i = 0; i < values.size() / 2; i++)
```
S-Dafarra, line 355: Why the /2?

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -329,61 +330,40 @@ bool GloveWearableImpl::getFingertipPoseValues(Eigen::MatrixXd& values)
...
-            << "size of the vibrotactile feedback vector is not equal to the size of the default "
-               "vibrotactile feedback size.";
+        yError() << m_logPrefix
+                 << "size of the haptic feedback vector is not equal to the size of the default "
+                    "haptic feedback size.";
     }
+    wearable::msg::WearableActuatorCommand& wearableActuatorCommand
+        = m_iWearActuatorPort.prepare();
+    wearableActuatorCommand.forceValue.resize(m_numForceFeedback, 0);
+    wearableActuatorCommand.vibroTactileValue.resize(m_numVibrotactileFeedback, 0);
```
S-Dafarra, line 352:
```suggestion
    wearableActuatorCommand.forceValue.resize(m_numForceFeedback);
    wearableActuatorCommand.vibroTactileValue.resize(m_numVibrotactileFeedback);
```
The values will be set down below anyhow

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -329,61 +330,40 @@ bool GloveWearableImpl::getFingertipPoseValues(Eigen::MatrixXd& values)
...
+    wearableActuatorCommand.duration.resize(m_numForceFeedback + m_numVibrotactileFeedback, 0);
```
S-Dafarra, line 353: Is this duration field needed?

modules/HapticGlove_module/include/GloveControlHelper.hpp
EhsanRanjbari: Yes, makes more sense. Done!

modules/HapticGlove_module/src/GloveWearable.cpp
EhsanRanjbari: because first half is for force and the second half is for the vibrotactile feedback.

modules/HapticGlove_module/src/GloveWearable.cpp
EhsanRanjbari: Yes. I will remove them.

modules/HapticGlove_module/src/GloveWearable.cpp
S-Dafarra: Ah I see, then this is a duplicate of https://github.com/robotology/walking-teleoperation/pull/139#discussion_r1521749794

modules/HapticGlove_module/src/GloveWearable.cpp
EhsanRanjbari: Not actually. Can be removed! So far, it is always zero.

modules/HapticGlove_module/src/GloveControlHelper.cpp
EhsanRanjbari: It is possible. But can you please explain more on why do you think it is better?

S-Dafarra: It is better to separate different data. As it is right now the receiver also needs to know that half of it is force, the other half is haptic. It is very error-prone, also because you might have a different number of haptic and force feedback values

EhsanRanjbari: I tried to separate the vectors with this commit: https://github.com/robotology/walking-teleoperation/pull/139/commits/5c692d3f3db0904230bef887230277d846a7ebaa

_Note: will need to test it though!_

mebbaid (changes requested):

modules/HapticGlove_module/include/GloveControlHelper.hpp
```diff
@@ -93,6 +94,12 @@ class HapticGlove::GloveControlHelper
      */
     bool getHandJointAngles(std::vector<double>& jointAngles);
 
+    /**
+     * Set the desired haptic feedback reference to the user fingertip
+     * @return true/false in case of success/failure
+     */
+    bool setFingertipHapticFeedbackReferences();
```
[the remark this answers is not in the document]
EhsanRanjbari: > I was expecting that this method takes a parameter being desiredHapticValue or something similar.

Actually, it returns another method that takes desiredHapticValue. 

https://github.com/EhsanRanjbari/walking-teleoperation/blob/5c692d3f3db0904230bef887230277d846a7ebaa/modules/HapticGlove_module/src/GloveControlHelper.cpp#L206-L210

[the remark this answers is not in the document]
mebbaid: This then leads to the comment below. that we have `m_desiredHapticValues` and we are not using that in the set method. Instead we always separate Force and vibrotactile.

mebbaid:
Could you rebase on top of master @EhsanRanjbari. Thanks

mebbaid (approved):

modules/HapticGlove_module/src/GloveWearable.cpp
```diff
@@ -403,15 +375,6 @@ bool GloveWearableImpl::setPalmVibrotactileValue(const int& value)
...
 
-    wearable::msg::WearableActuatorCommand& wearableActuatorCommand = m_iWearActuatorPort.prepare();
-
-    wearableActuatorCommand.value = value;
-    wearableActuatorCommand.info.name = m_wearablePrefix + wearable::actuator::IHaptic::getPrefix()
-                                        + m_handLinkName + "::VibroTactileFeedback";
-    wearableActuatorCommand.info.type = wearable::msg::ActuatorType::HAPTIC;
-    wearableActuatorCommand.duration = 0;
-
-    m_iWearActuatorPort.write(false);
```
S-Dafarra, line 414: It the method ``setPalmVibrotactileValue`` still useful then?

S-Dafarra:
@EhsanRanjbari, was it intended to close this?

EhsanRanjbari:
> @EhsanRanjbari, was it intended to close this?

Oops, I deleted this by mistake. Fortunately I had it locally and I was able to recreate it. I opened a new branch (instead of a fork) and related PR here: https://github.com/robotology/walking-teleoperation/pull/143

PR #146, opened 2024-08-01
Reduce compilation warnings

S-Dafarra:
Similar to https://github.com/robotology/walking-controllers/pull/181, I tried to reduce the compilation warnings

It needs #144 to be merged first.

PR #148, opened 2024-09-10
Added explicit connection to IWear remapper in HapticGloveModule

S-Dafarra:
It has https://github.com/robotology/wearables/pull/211 as runtime dependency, but it is backward compatible

Requires https://github.com/robotology/walking-teleoperation/pull/146 to be merged first


S-Dafarra:
I have tested this on the robot and works fine

mebbaid (approved):

